PR #139, opened 2014-09-20
Adds check to useColors for worker contexts

nikcorg:
When bundling `debug` using `browserify` the browser version is used, and the `useColors` check crashes when in a ServiceWorker, because `document`, `window` and `navigator` are all `undefined`.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -46,12 +46,14 @@ exports.colors = [
 
 function useColors() {
   // is webkit? http://stackoverflow.com/a/16459606/376773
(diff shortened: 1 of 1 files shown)


nikcorg:
@TooTallNate I modified the PR slightly. It now adds checks for `WebWorker`, `SharedWorker` and `ServiceWorker` contexts. They're still a bit special cased and possibly a bit unnecessarily explicit, but should do no harm.

thebigredgeek:
Closed in lieu of https://github.com/visionmedia/debug/pull/298/files

PR #140, opened 2014-09-24
Process all arguments to debug call instead of only the first.

### LINKED ISSUE #132: Passing several arguments to debug() looks weird
I am debugging SQL queries like:

``` js
var debug = require('debug')('pg-bricks');
debug(query, params);
```

And I got query grayed, but not params:

pg-bricks-logging

woutgg:
This would fix #132.

### DIFF
--- a/debug.js
+++ b/debug.js
@@ -87,10 +87,12 @@ function debug(namespace) {
 
     args[0] = exports.coerce(args[0]);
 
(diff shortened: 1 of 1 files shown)


agirorn:
Can't debug detect if formatting is not being used and then print out all the arguments and when formatting is use can't debug print out all the leftover arguments ?

Issue #150, opened 2014-10-24
debug.disable() does not disable?


swook:
```
process.env.DEBUG="*";
var debug = require("debug");
var log = debug("app:log");
log("Should show");
debug.disable();
log("Shouldn't show");
```

The above snippet does not work as expected.

The disable function should remove namespaces instead of trying to enable nothing.
https://github.com/visionmedia/debug/blob/master/debug.js#L159

CumpsD:
`debug.enable(null);` does a real disable, removing the localStorage in the browser version, while `debug.disable()` simply sets it to an empty string.

ibc:
`null == ''` returns false, so @CumpsD is right and this is a bug in the code. `debug.disable()` should call to `exports.enable(null)` rather than `exports.enable('')`.

Anyhow, calling to `exports.enable(null)` will just disable the debug for future created debug instances (that is what this issue is about).

ibc:
Another problem is that calling  `debug.enable('stuff')` enables future debug instances whose prefix matches "stuff", but the DEBUG variable (or the "debug" item in LocalStorage in the browser) is overridden with just "stuff" (this is, just the last prefix given to `debug.enable()`).

This is inconsistent.

thebigredgeek:
https://github.com/visionmedia/debug/issues/370 < V3 should solve this

PR #151, opened 2014-10-25
Fix debug.disable()

swook:
Addresses issue #150

`exports.disable()` now resets the array `exports.names` to actually disable future debug instances.

### DIFF
--- a/debug.js
+++ b/debug.js
@@ -157,7 +157,7 @@ function enable(namespaces) {
  */
 
 function disable() {
-  exports.enable('');
+  exports.names = [];
 }
 
 /**


thebigredgeek:
@TooTallNate @swook we should probably figure out the best way to move forward on this.  Do we want to extend the API or handle the skips reset in one of the existing functions?

thebigredgeek:
@swook would love to get this merged by tonight so I can release tomorrow with 2.4.0

swook:
My opinion on this matter has not changed since my last comment.

Please refer to #150 for other comments.

thebigredgeek:
closing this as it doesn't sound like there is alignment on direction :(

PR #163, opened 2014-12-14
Add configuration using .debugrc file

rkgarcia:
It's working for node


TooTallNate:
Thanks, but no thanks to be honest. Having > 1 configuration mechanism is just feature bloat in my opinion.

You could keep this route by writing a program that parses `.debugrc` and sets the `DEBUG` env variable and run that in your `.bashrc` file, and you would basically get the same result.

Thanks again!

PR #164, opened 2014-12-19
Option for using / not-using space delimited namespaces.

dylan-baskind:
Personally I find logger names with spaces more readable 'Something Long Here"  rather than "Something:LongHere".

Not very fully realised; but hopefully something useful if / when updates to debug.js happen.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
-  "name": "debug",
-  "version": "2.1.0",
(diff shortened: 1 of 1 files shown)


dylan-baskind:
No worries - (yah figured it was more of a personal taste / much-of-a-muchness thing ).

Thanks for debug.js :)

PR #174, opened 2015-01-24
changed bower package name

andrzejdus:
Slashes are not allowed in bower component name (https://github.com/bower/bower.json-spec).

Slashes also cause practical problems (at least on OSX). For example, bower installs component files without any problems, but on update ends up with error saying that package is not installed.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,12 +1,12 @@
 {
-  "name": "visionmedia/debug",
+  "name": "visionmedia-debug",
   "main": "dist/debug.js",
   "version": "2.1.1",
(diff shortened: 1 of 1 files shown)


jitowix:
Please merge this asap, I can't install the package atm.

TooTallNate:
Would simply `debug` work?

jitowix:
I think there is already a package with this name.

jitowix:
When I do `bower install visionmedia-debug` it doesn't find the package. Is it normal?

imposibrus:
@jitowix Need to create a new release first, and, maybe re-register package in Bower.

PR #189, opened 2015-03-26
Makefile: add a `distclean` target

### LINKED ISSUE #162: don't remove node_modules in clean target
One way to do cross compilation of nodejs apps is to "make clean" across the tree, zip up the sources and run nmp rebuild for the target. This assumes that "make clean" would leave the npm tree intact.

You should be able to use "npm install --force" to reinstall, right?

stephenmathieson:
This patch prevents the `clean` target from removing node_modules and
adds a `distclean` target instead.

Fixes #162.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ all: dist/debug.js
 install: node_modules
 
 clean:
-	@rm -rf node_modules dist
+	@rm -rf dist
 
 dist:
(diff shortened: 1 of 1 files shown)


TooTallNate:
Thanks @stephenmathieson!

stephenmathieson:
np

PR #195, opened 2015-04-08
Catch JSON stringily errors.

jalleyne:
There is a chance that there may be circular references in the Object which will cause the debug message to throw an exception. If the exception is not handled on the application level this will cause node to crash. As a debug library, errors in here should not have such a serious impact on the application.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -59,7 +59,11 @@ function useColors() {
  */
 
 exports.formatters.j = function(v) {
-  return JSON.stringify(v);
+  try {
+    return JSON.stringify(v);
+  }catch( err){
+    return '[UnexpectedJSONParseError]: ' + err.message;
+  }
 };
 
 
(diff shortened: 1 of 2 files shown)


stephenmathieson:
imo, this falls out of scope for this module.. see what nate says though

jalleyne:
Completely agree that argument validation is out of scope, but dont think its a good idea that debug code can throw unexpected errors. This is sort of an odd case.

ronkorving:
Since it's debug doing the stringify operation, it's its responsibility not to throw. Node.js's own console.log deals with this gracefully by mentioning circular references explicitly. I think that's the only correct way to deal with this.

jalleyne:
Any updates/ further thoughts on this?

world:
I agree that a debug module should not be able to crash node. I also would like to know if debug is dead (we haven't heard from Nate in way too long and there are lots of legit pull requests from 2015. Should we fork and rename and move this thing forward?)

TooTallNate:
👍

thebigredgeek:
Looks good to me!  Merging

PR #197, opened 2015-04-19
Make debug(value) more functional by returning value

MichaelJCole:
PR to have debug('value') return 'value'.  This makes debug() more functional and supports this use case:

```
  var h = someHelperThatMakesHttpRequests();

  it('should create a new user', function(done) {
    h.post(h.uri + '/sign-up')
    .send(debug(h.randomUserData()))  // Here I want to debug the return value of this function call
    .end(function(err, res){
      debug(res.body);
      res.status.should.equal(200);
      done(err);
    });
  });
```

Thanks!

### DIFF
--- a/debug.js
+++ b/debug.js
@@ -62,7 +62,8 @@ function selectColor() {
 function debug(namespace) {
 
   // define the `disabled` version
-  function disabled() {
+  function disabled(data) {
+    return data;  // Return first parameter.  debug('sam i am') === 'sam i am')
   }
   disabled.enabled = false;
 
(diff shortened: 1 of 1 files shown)


stephenmathieson:
-1 from me. this is a breaking change.

I've got stuff like this everywhere, which would no longer work:

```js
if (whatever())
  doStuff()
else
  doOtherStuff()

function whatever(){
  if (something)
    return debug('blah')
  devug('cool')
  return true
}
```

MichaelJCole:
Why would you specifically return 'undefined' in whatever() and then test for it?  It seems like a bug in your code.

The example shows how the API is changing for an edge case and seems a bit contrived - which is fine.

My point is those bugs are caused by [the API returning 'undefined' instead of a value](https://wiki.haskell.org/Functional_programming).

stephenmathieson:
what? how is returning a falsey value a bug?

thebigredgeek:
@MichaelJCole I think it's pretty common practice to return from a function call to end a block.  Closing this for now

PR #203, opened 2015-05-13
Fix #150

### LINKED ISSUE #150: debug.disable() does not disable?

mgesmundo:


thebigredgeek:
@mgesmundo can you update with latest code please?  Then we can discuss merging this in

thebigredgeek:
@mgesmundo thanks for the PR.  Can you point out exactly the code that applied the fix?  We might want to move that into a separate PR, or you might want to revert the code style changes you applied in addition to the fix? Sorry to be a stickler, but until we have tests I am trying to avoid lots of code style changes :p

thebigredgeek:
Closing due to crazy old and unresponsive

PR #204, opened 2015-05-19
Allow debug to be used within a web worker thread

JesseFarebro:
When trying to use the debug module within the browser in the context of a web worker, document and window aren't available, therefore it will throw an exception. Checking for document and window before using colors seems to be the only place where these checks are performed before initialization.


nikcorg:
This is pretty similar to #139.

Issue #154, opened 2014-11-07
how to avoid the tedious initialization?


stablum:
also posted here: http://stackoverflow.com/questions/26777433/node-js-debug-module-how-to-avoid-this-tedious-initialization

I am wondering if there is any way to prevent this tedious and overly redundant inizialization in each module:

```
require('debug').enable('module-name:log module-name:ERROR');
var log = require('debug')('module-name:log');
var error = require('debug')('module-name:ERROR');
```

As you can see it's redundant on two different levels: I need to "enable" the loggers before instantiating them, sending the same name twice as strings, and I need to write the module name. Is there any way to do this automagically?

dmarcelino:
Hi @stablum, I've just replied to your stackoverflow question. If you are using `debug` outside the browser you can use [debug-logger](https://github.com/appscot/debug-logger) to lazily instantiate multiple instances of `debug` to log at different levels as you would do with console. It's essentially a wrapper for `debug` with no other dependencies.

Example:
```javascript
var log = require('debug-logger')('myapp');

log.trace("I'm a trace output");
log.debug("I'm a debug output");
log.log("I'm a log output");
log.info("I'm an info output");
log.warn("I'm a warn output");
log.error("I'm an error output");
```

would print:

``` shell
myapp:trace I'm a trace output +0ms
myapp:debug I'm a debug output +2ms
myapp:log   I'm a log output +0ms
myapp:info  I'm an info output +0ms
myapp:warn  I'm a warn output +1ms
myapp:error I'm an error output +0ms
```

More info at the project page,
Regards

ashwell:
I know this is a year later, but I just wrote this to help facilitate creating debug instances for stdout and stderr...

```js

const debug = require( 'debug' );

module.exports = function( namespace ) {
  const
    log = debug( `${ namespace }:log` ),
    error = debug( `${ namespace }:error` );

  log.log = console.log.bind( console );

  return { log, error };
};
```

thebigredgeek:
Closing this for now.  Feel free to discuss further the V3 API RFC issue: https://github.com/visionmedia/debug/issues/370

PR #209, opened 2015-07-02
Add optional dynamic debug instances (react on enable() and disable() once created)

ibc:
First of all: this PR is 100% **compliant** with the current design and adds **zero** overhead to it. This is, no changes must be done in apps and libraries already using the _debug_ module, and the performance is the same as before.

This PR adds "dynamic debug instances" which are created by passing `true` as the second parameter to the module exported function:

``` js
var debug = require('debug');
var applog = debug('myApp', true);
```

Those instances **do properly** react if `debug.enable(xxxx)` or `debug.disable()` is called after they are created.

An [usage example](https://github.com/ibc/debug/blob/master/example/dynamic.js) is provided in the _example/_ folder.

The PR also resets `exports.names` and `exports.skips` when `enable()` or `disable()` is called (rationale [here](https://github.com/visionmedia/debug/issues/150#issuecomment-60466454)).

Related issues:
- https://github.com/visionmedia/debug/issues/150
- https://github.com/visionmedia/debug/issues/154

### DIFF
--- a/debug.js
+++ b/debug.js
@@ -29,6 +29,18 @@ exports.skips = [];
 exports.formatters = {};
 
 /**
+ * Container for dynamic debug instances.
+ */
+
+exports.dynamics = {};
+
+/**
+ * Dynamic debug instances counter.
+ */
+
+var dynamicCounter = 0;
+
+/**
  * Previously assigned color.
  */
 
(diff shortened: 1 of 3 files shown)


leftieFriele:
+1 on this PR, turning on / off on the fly would be great for browser applications

ibc:
@TooTallNate any comment about this?

ibc:
Author, thanks a lot for your indolence.

bragma:
+1 please merge this!

wision:
@TooTallNate any chance of merging this feature?

techjeffharris:
:+1: Yes, please!

debug.js
```diff
@@ -55,74 +67,145 @@ function selectColor() {
...
-      if ('function' === typeof formatter) {
-        var val = args[index];
-        match = formatter.call(self, val);
-
-        // now we need to remove `args[index]` since it's inlined in the `format`
-        args.splice(index, 1);
-        index--;
+  // Return a dynamic debug instance.
+  else {
+    function dynamic() {
```
stephenmathieson: -1 for all this copypasta

ibc: What do you mean??

stephenmathieson: the `dynamic` function is nearly identical to the `enabled` function

ibc: Right. We can talk about code optimization if the PR is at least considered by the author, which sai nothing about it.

roccomuso:
Would it be referred to #275 ?

thebigredgeek:
@TooTallNate thoughts here?  Not sure if we want to go this direction with the API

thebigredgeek:
https://github.com/visionmedia/debug/pull/156

thebigredgeek:
https://github.com/visionmedia/debug/issues/326

TooTallNate:
A few comments:

 - I'm not a fan of the new API surface area. Why not all of them just be dynamic? And forget the ugly `, true` second argument to debug (boolean trap waiting to happen).
 - My PR in #327 (specifically this commit f0e8a97e92a6ac6edf6f46c039efb4b5e22c049b) makes individual debug instances "hot-enableable" (that is, if you toggle the `enabled` flag on a debug instance it works as expected). Perhaps use that as a base instead of the "copy pasta".
 - You have to explicitly release dynamic debug instances otherwise `exports.dynamics` is a memory leak (plus it's an API, not change, but addition).

TooTallNate:
Honestly that last point is the main reason a feature like this is not in debug right now. I don't want to have to maintain a list of all the debug instances in memory, which just grows and grows as more modules are required.

thebigredgeek:
@TooTallNate agreed.  Closing this for now.

tehsenaus:
I don't see how this would cause a leak - the number of modules which can be require()'d is bounded, is it not?

ibc:
> I don't see how this would cause a leak - the number of modules which can be require()'d is bounded, is it not?

If you want to dynamicall enable/disable a debug instance you must store it somewhere. If you create tons of debug instances and never destroy them, they will in fact leak. Easy.

That's why in my PR I added a [debug.release()](https://github.com/visionmedia/debug/pull/209/files#diff-9fd8b683aa0fd2f9b1cb7245c42c96efR202) method.

But, the current implementation of this module is too not handle debug instances. Instead they just exist while they are in the scope, and they are garbage collected if not like any variable.

tehsenaus:
If you're creating debug instances on the fly and not releasing them, then sure. But is the library being used in that way? I'd bet 99.99% of cases users just ```require('debug')('blah:blah')``` at the top of their module. Then by the same logic, require() leaks memory!

This would be sooo useful for debugging in production, it seems a real shame to pass on for such an esoteric use case. If I were to wrap the library in userland, i'd only get the functionality for my application (not submodules like express).

ibc:
> If you're creating debug instances on the fly and not releasing them, then sure. But is the library being used in that way? I'd bet 99.99% of cases users just require('debug')('blah:blah') at the top of their module.

Sure. However sometimes I create dynamic debug instances in class constructors so their label include the `id` or whatever.

tehsenaus:
That doesn't seem too efficient, it would be running lots of regex checks on your class instantiation?

Why not just ```log('%s: ...', this.id)```?

thebigredgeek:
The reason this leaks is because internal state within the debug lib is required.  If you call debug over and over in it's current state it simply returns a function.  This PR requires it to keep an ever growing amount of bookkeeping.  A lot of people create debug instances in class constructors as mentioned before.  This would be a breaking change that would manifest itself as a server crash once ram filled up.  I get that it's useful, but that doesn't necessarily mean it's doable.  There are always tradeoffs

ibc:
> That doesn't seem too efficient, it would be running lots of regex checks on your class instantiation?

I agree, but that's up to the app developer.

> This PR requires it to keep an ever growing amount of bookkeeping.

Not true. This PR does **NOT** change the current/default behavior. If the user wants manageable debug instances it must create them by passing `true` as second parameter:

```js
var dyndebug = require('debug')('myApp', true);
```

I understand that such an API (passing just `true` as second argument) is not too cool, but you got the idea.

tehsenaus:
require()s are synchronous, right? So if one were to enable bookkeeping initially on startup, then disable bookkeeping in a process.nextTick() (so dynamic debug instances would not be tracked), this would achieve the desired effect with no leaks.

Food for thought...?

Odrinwhite:
I would argue with the API for enabling that feature, but I would definitely love to see it in next major release, enabled by default.

thebigredgeek:
Check out the 3.0 conversation issue.  It's something that will be added.  However, I don't think we want to resort to hackery to make it work.  We should probably just save enables/disabled state on a boolean within the factory closure as that won't leak

tehsenaus:
I need this feature, so I created a module which enables it with no API changes:
https://github.com/tehsenaus/hot-debug

It patches the debug module, so any sub-dependencies of your project (e.g. express) will get the enable/disable functionality. 

It'd be great to see this feature natively in debug v3!

Yalhu:
There are some questions that I use debug on Browser. 
1. Does `debug.disable('debug:debug1')` disable all the debug? On my test ,it was.
2. Does `debug.enable()` must have a argument. If there is no argument, it will disable debug.
3. If I disable debug on the script start up, on console `debug.enable('debug:debug2')` doesn't work. It will works again if I use `debug.enable('debug:*')`. (I use these API on chrome console).

### steps
```
localStorage.setItem('debug','debug:*');
var d1=debug('debug:debug1');
var d2=debug('debug:debug2');
debug.disable('debug:debug2'); // only debug2,but d1 not work
d1('test d1 ....');
d2('test d2 ...');
```
on chrome console:
1. there is no debug.
2. do `debug.enable('debug:debug1') d1('can work')` worked.
3. do `debug.enable('debug:debug2') d2('can work')` not worked.
4. do `debug.enable('debug:*') d2('can work')` worked.
5. do `debug.enable() d1('can work') d2('can work')` no debug.

PR #212, opened 2015-08-01
expose debug as default export

bucaran:
+ Similar to [tj/co](https://github.com/tj/co/blob/master/index.js#L12)
+ Allow to re-export debug from another module in ES6

Without this fix, it is troublesome to re-export debug from a module. Basically, the following will not compile / `debug` will not be correctly exported as it should:

```js
// my utilities
export myLog () {}
export * from "debug" // just re-export debug 
...
```

### DIFF
--- a/debug.js
+++ b/debug.js
@@ -6,7 +6,7 @@
  * Expose `debug()` as the module.
  */
 
-exports = module.exports = debug;
+exports = module.exports = debug.debug = debug;
 exports.coerce = coerce;
 exports.disable = disable;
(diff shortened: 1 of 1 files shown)


bucaran:
@TooTallNate @tj Do you think we can have this merged? 

Best.

bucaran:
@TooTallNate Can we can have this merged?

bucaran:
Thanks

101100:
Any plans to release this on `npm` soon?

PR #215, opened 2015-08-25
move ms to the last place

gorangajic:
when you call debug with more than one param ms diff goes to the second place

```javascript
var debug = require('debug')('OK');
debug('awesome', 'debug', 'message');  
```
  before
```OK awesome +0ms debug message```
  
  after
 ```OK awesome debug message +0ms```

### DIFF
--- a/node.js
+++ b/node.js
@@ -80,17 +80,21 @@ exports.formatters.o = function(v) {
  */
 
 function formatArgs() {
-  var args = arguments;
+  var args = [];
+  var len = arguments.length;
   var useColors = this.useColors;
   var name = this.namespace;
+  for (var i = 0; i < len; i++) {
+    args.push(arguments[i]);
+  }
 
   if (useColors) {
     var c = this.color;
 
     args[0] = '  \u001b[3' + c + ';1m' + name + ' '
       + '\u001b[0m'
-      + args[0] + '\u001b[3' + c + 'm'
-      + ' +' + exports.humanize(this.diff) + '\u001b[0m';
+      + args[0];
+    args.push('\u001b[3' + c + 'm+' + exports.humanize(this.diff) + '\u001b[0m');
   } else {
     args[0] = new Date().toUTCString()
       + ' ' + name + ' ' + args[0];

### CODE REVIEW

**Comment by steelbrain** [comment]
:+1:

**Comment by TrejGun** [comment]
:+1:

**Comment by mbroadst** [comment]
:+1:

**Comment by steelbrain** [comment]
I wonder what exactly is stopping us from shipping this.

cc @TooTallNate

**Comment by mbroadst** [comment]
@steelbrain it may be worth noting that arguments are potentially leaked here, see: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments

**Comment by steelbrain** [comment]
@mbroadst I've seen this used in the node core itself (Although they use `Array.from` instead of `Array.prototype.slice.call(arguments)` but the concept remains the same), that being said I don't think the performance hit will be significant (unless you have a benchmark proving otherwise)

**Comment by gorangajic** [comment]
not sure it's worth it but I have updated pull request, so now arguments are not leaked

didn't know about `Array.from` :+1:

**Comment by mbroadst** [comment]
@steelbrain Array.from I believe will make a new copy of the array, and therefore not leak. I'm not really being a stickler here I was just speculating as to why it hasn't been merged yet :) not to mention the performance hit would be negligible considering debug should be a no-op if you haven't enabled debugging for the given category

@gorangajic Also I'm pretty sure we can't use Array.from here since its ES6 and this needs to target older versions of node?

**Comment by mbroadst** [comment]
@gorangajic oops, totally wrote that without looking at your code update please ignore the noise!

File: node.js [hunk 1]
```diff
@@ -80,17 +80,22 @@ exports.formatters.o = function(v) {
  */
 
 function formatArgs() {
-  var args = arguments;
+  var args = [];
+  var len = arguments.length;
+  var i = 0;
   var useColors = this.useColors;
   var name = this.namespace;
+  for ( ; i < len; i++) {
```
steelbrain: Even though I don't maintain the project, I would still like to hear why `i` is declared above and not here :)

mbroadst: personal preference I'm sure, hoisting and all that

mbroadst:
@TooTallNate any updates here?

world:
this is a great bug fix. I have forked from @TooTallNate who may have abandoned debug from what I can tell. Thank you @gorangajic

mbroadst:
cc @TooTallNate ?

thebigredgeek:
This could be a breaking change.  A lot of people rely on the existing format when parsing / aggregating logs.  Is there any real reason that you would want to change the format here?

node.js
thebigredgeek: can you inline the `i` in the for loop please?

thebigredgeek:
I am reviewing.  Once change is applied I'll gladly merge this

gorangajic:
@thebigredgeek updated pull request

thebigredgeek:
@gorangajic merging this.  Thanks for the contribution

PR #224, opened 2015-09-16
Easier browser instructions

matthewmueller:
Learned this trick from @rauchg. much easier and doesn't require a globally exposed debug.


TooTallNate:
Lol, damn. That's what I've always done. Bad outdated docs! Thanks Matt!

PR #225, opened 2015-09-21
browser-supports-log-styles

gr2m:
I need a check if a browser supports console log styles in another library, so I created a tiny npm package [browser-supports-log-styles](https://www.npmjs.com/package/browser-supports-log-styles) for that, and thought you might want to use that as well, so we can keep that up-to-date together?

Let me know if you want any changes in that, it's your work after all, I don't mean to take any credit for it

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -10,7 +10,7 @@ exports.log = log;
 exports.formatArgs = formatArgs;
 exports.save = save;
 exports.load = load;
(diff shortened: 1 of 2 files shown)


thebigredgeek:
Trying to keep our library as minimal as possible.  Thanks anyway!  Closing PR

PR #241, opened 2015-12-13
Check for web workers, fixes #240

### LINKED ISSUE #240: Presence of "document" can't be assumed for web workers
Currently `debug` fails in a web worker, because `document` is not available.

Background: https://github.com/pouchdb/pouchdb/issues/4612

nolanlawson:
It should be sufficient to check for a global `window` object. If the user is bonkers and actually declares a global `window` inside of their web worker... then well, we can't help them. The alternative would be to check for all three of `window`, `document`, and `navigator`, which I think is excessive.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -38,6 +38,11 @@ exports.colors = [
  */
 
 function useColors() {
+  // check we aren't in a web worker
+  if (typeof window === 'undefined') {
+    return false;
+  }
+  
   // is webkit? http://stackoverflow.com/a/16459606/376773
   return ('WebkitAppearance' in document.documentElement.style) ||
     // is firebug? http://stackoverflow.com/a/398120/376773


epeli:
Except that you can use colors in Chrome Web Worker :)

I also needed to add a fake local storage implementation in order to get actually get it logging: https://github.com/epeli/debug/commit/f5cb7d94d93d97c8ee9330dc09e8142b27275e20

FYI: I'm hacking here because I'm trying to use debug with React Native and Chrome debugging.

grimen:
Any update on this? I just realized `debug` is not working in Service Workers nor Web Workers which should be considered a bug, this definitely should be merged asap...

botverse:
Also this won't fix #286, #303, #307

thebigredgeek:
Closed in lieu of https://github.com/visionmedia/debug/pull/298 (already released)

PR #243, opened 2015-12-16
Add a note for PowerShell users.

unreadable:


TooTallNate:
Thanks!

Readme.md
```diff
@@ -58,6 +58,12 @@ setInterval(function(){
  ```cmd
  set DEBUG=*,-not_this
  ```
+ 
+ Note that PowerShell using different syntax to set environment variables.
```
stephenmathieson: ocd: should be "uses a"

PR #244, opened 2015-12-23
Make timers namespace specific and add fast Always Enabled support 

gtjoseph:


thebigredgeek:
@TooTallNate thoughts?

freewil:
Name-spacing the timers makes sense to me. In the past, I found the global timer behavior very non-intuitive when I had multiple workers with different namespaces logging.

debug.js
```diff
@@ -38,7 +38,7 @@ var prevColor = 0;
  * Previous log timestamp.
  */
 
-var prevTime;
+var prevTime = {};
```
thebigredgeek: In this case, the previous time should be held inside of the debug closure.  That way the value is encapsulated by the log instance.  Technically, this line is a leak

thebigredgeek:
@gtjoseph bump?

gtjoseph:
Wasn't paying attention, sorry. :)

I'll get everything fixed up tonight.

gtjoseph:
Messed up.  Splitting the pull request into 2 separate ones.

PR #211, opened 2015-08-01
Disable colors when the browser is PhantomJS.

ahfarmer:
Because PhantomJS is WebKit, `debug` thinks it supports colors, but PhantomJS logs to the terminal and the output comes out like this:

```
 '%cManager %cstyle%c +20ms', 'color: lightseagreen', 'color: inherit', 'color: lightseagreen'
```

After this PR, colors are disabled and the output looks OK:

```
Manager style +20ms
```

If you want any changes to this PR before merging just let me know! 😸

PR #247, opened 2015-12-31
added parameter to useColors to allow setting an override

### LINKED ISSUE #246: useColors should take an override parameter and save it
I'm using debug in a Qt-WebKit application and debug falsely thinks it can handle color.

It would be nice if users could call useColors with a parameter to set an override.

jandrieu:
Resolves #246

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -37,8 +37,15 @@ exports.colors = [
  * TODO: add a `localStorage` variable to explicitly enable/disable colors
  */
 
-function useColors() {
-  // is webkit? http://stackoverflow.com/a/16459606/376773
+function useColors(force) {
+  if (typeof force != 'undefined') {
+    this.forceColors = !!force;
+  }
+  
+  if ('forceColors' in this)
+    return this.forceColors;
+    
+    // is webkit? http://stackoverflow.com/a/16459606/376773
   return ('WebkitAppearance' in document.documentElement.style) ||
     // is firebug? http://stackoverflow.com/a/398120/376773
     (window.console && (console.firebug || (console.exception && console.table))) ||


jandrieu:
Hmmm... after I made that change, I saw the TODO to add a 'localStorage' variable.  Maybe that's a better option...  This fix works for my use case, however. Maybe you'll like it.

poshaughnessy:
We're trying to use debug in a Cordova application, using a Cordova plugin to output console.logs as native logs. It would be good to force-disable the colours, since our native app logs come out looking like "%c this %c".

I would vote for making it a simple parameter rather than having to store a value in localStorage!

thebigredgeek:
@jandrieu I think this would possibly work better as an environment variable on the server.  On the client, we probably also want to respect `process.env` namespace (as is frequently done with webpack)

thebigredgeek:
@TooTallNate thoughts?  Using an environment variable like `DEBUG_COLORS=false` might be a good solution here.  Also solves https://github.com/visionmedia/debug/pull/211

thebigredgeek:
https://github.com/visionmedia/debug/issues/314

thebigredgeek:
https://github.com/visionmedia/debug/issues/309

thebigredgeek:
Need to update code @jandrieu

jandrieu:
I don't think the environment variable solves my use case.  I use debug in a custom browser, with my code bundled with browserified. The browser can't handle the colors and I don't believe browserify has a way to pass environment variables to debug.

This patch allows the javascript within the page to turn colors on/off programmatically. My custom browser provides enough information that the script can detect that it is running in a normal browser or my custom browser, then the script calls use colors to set true/false.

thebigredgeek:
@jandrieu check out middleware proposal here.  would love your input so that v3 supports your needs with this.

https://github.com/visionmedia/debug/pull/247

thebigredgeek:
closing for now

thebigredgeek:
https://github.com/visionmedia/debug/issues/370 Middleware should allow for this

gpbl:
@thebigredgeek how does it work? I can’t see any docs about middlewares :(

PR #251, opened 2016-01-17
Revert "npmignore: don't publish the `dist` dir"

### LINKED ISSUE #242: Why not include 'dist' folder in npm package?
When i use webpack to compile file, console show these error message.

```
ERROR in debug (bower component)
Module not found: Error: Cannot resolve 'file' or 'directory' ./dist/debug.js in /Users/stevenjlho/Work/v4s/htdocs/node_modules/debug
@ debug (bower component) 1:17-43
```

Npm package version is 2.2.0.

nhaesler:
This reverts commit 43b0d51c21134e86014f0ba007dfdabc6b21a35e.

This fixes #242

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -3,4 +3,3 @@ test
 examples
 example
 *.sock
-dist


TooTallNate:
To clarify, are you using webpack, or bower, or other?

nhaesler:
I'm using npm.

nhaesler:
To clarify this a bit more, this is needed because we (like many others) use npm as a front-end package manager without any module loader at the moment. Therefore we need the dist/debug.js, which includes a module loader at the top of the file.
If you're using webpack, you can just use the browser.js.
If you're using bower, you get the dist/debug.js, because it ignores the .npmignore file.

cema-sp:
+1

PeterUstinox:
+1

keik:
+1

Oceanswave:
+1

addaleax:
bump @TooTallNate ?

thebigredgeek:
To be clear, how are you using node_modules for client side packages exactly?  Are you copying them at build time?  It seems like usually a module loader like webpack or browserify is probably a better approach...

nhaesler:
@thebigredgeek At the time we just replaced bower with npm and served node_modules/ like we did with bower_components/ before  like I explained in [#issuecomment-172400205](#issuecomment-172400205). It worked with bower before, because it ignores the .npmignore and therefore gets the dist/debug.js. Not every little project wants to use a module loader or webpack, so accepting this pull request would be really helpful. Especially since I really don't see the reason for not including the distribution file in the actual release on npm. If you look at all the comments on this pull request and on #242 it really seems to be a problem for a lot of people. Could you maybe explain why you're so hesitant to accept this change?

thebigredgeek:
@nhaesler yeah I understand that it might be helpful, but it unfortunately more than triples the gzip size on NPM which is not really something that we really want to do if we can avoid it.  One of the major reasons a lot of libraries use debug is the small footprint.

I am also having trouble understanding why you'd want to avoid using webpack in this case.  You should be able to set up webpack for most projects in under 5 minutes, and it sounds like it would greatly simplify your deployment.  I am not sure if exposing node_modules via static middleware or the like is a good idea, though I suppose that is a different conversation :p.  If you want to keep things simple, I think bower is designed for exactly this purpose.  I don't really know of a lot of people who use npm as a front-end package manager without something like browserify or webpack.

TooTallNate:
The `dist` dir isn't coming back.

If you're so set on using standalone modules in `<script>` tags (not recommended these days, as @thebigredgeek mentioned above), then you can use one of the packer CDNs, for example: https://wzrd.in/standalone/debug@2.2.0

PR #263, opened 2016-02-10
Print objects more clearly with more precise informations and avoid util.inspect

YouriT:
I propose my implementation of a replacer for util.inspect which is really poor in terms of help and colors and that can be very confusing and frustrating with large objects.

So what it basically does is replacing the use of utils.inspect for %o types and will beautifully output a clear information about an object without omitting to be deep when displaying.

### DIFF
--- a/debug.js
+++ b/debug.js
@@ -110,23 +110,25 @@ function debug(namespace) {
         });
 
         var stillNeedToParse = args[0].match(/%([a-z%])/g),
-            paramsToKeep = stillNeedToParse != void 0 ? stillNeedToParse.length : 0;
+            argsToKeep = 0;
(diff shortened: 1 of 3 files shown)


DevSide:
Could you take off all reformatted code ? It's really hard to see what you did.

TooTallNate:
I propose https://github.com/visionmedia/debug/pull/322 as an alternative. We shouldn't need to implement our own inspecting function when `util.inspect()` works just fine.

Thanks for the PR though!

jfoclpf:
Hi @TooTallNate, how can I _fully_ print an object without hidden information, that is without just showing `[Object]` for deep properties?

image

PR #264, opened 2016-02-10
Add child functionality to a debugger

thefotios:
This allows a developer to more easily create children loggers for larger modules.

This follows a similar convention to [`bunyan`](https://github.com/trentm/node-bunyan#logchild).

### DIFF
--- a/debug.js
+++ b/debug.js
@@ -122,6 +122,12 @@ function debug(namespace) {
 
   fn.namespace = namespace;
 
+  function createChild(childNamespace) {
+    return debug.debug(this.namespace + ':' + childNamespace);
+  }
+
+  fn.child = createChild;
+
   return fn;
 }
 
--- a/Readme.md
+++ b/Readme.md
@@ -58,9 +58,9 @@ setInterval(function(){
  ```cmd
  set DEBUG=*,-not_this
  ```
- 
+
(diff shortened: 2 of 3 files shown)


debug.js
```diff
@@ -122,6 +122,12 @@ function debug(namespace) {
 
   fn.namespace = namespace;
 
+  function createChild(childNamespace) {
+    return debug.debug(this.namespace + ':' + childNamespace);
```
DevSide: You can take off `this`

thebigredgeek:
Thanks for the PR!  Interesting idea.  For the sake of keeping the library minimalistic, can you provide a case where this is necessary rather than simply creating a new debug instance?  It's a cool feature, but it also becomes another moving part that we have to maintain.

TooTallNate:
I agree with @thebigredgeek. Out of scope for this library. Thanks for the PR though!

thebigredgeek:
@thefotios just as a note, you can also accomplish this on your own with some simple functional programming:

```js
function createChildDebugger (debugger, namespace) {
  return debug(debugger.namespace + ':' + namespace)
}

var debugger = debug('foo');

debugger('hello'); // => foo hello

var childDebugger = createChildDebugger(debugger, 'bar');

childDebugger('world'); // => foo:bar world
```

Issue #185, opened 2015-03-08
Clarification on object formatting


lgomez:
Hi!

I've searched for this but can't figure it out and have a feeling I'm just using it the wrong way. I read #125 but is there a way to output objects formatted like `debug("obj: %s", util.inspect(obj));` does but retain the coloring that one gets with `debug("obj: %o", response);`?

`debug("obj: %s", util.inspect(obj));` gives me a nested object with no coloring where `debug("obj: %o", response);` gives me the same object all on one single line but colored nicely.

Thank you!

lgomez:
I realize I can pass a colors option like so: `debug("obj: %s", util.inspect(obj, {colors: true}));` but I'm wondering if `debug("obj: %o", response);` can already achieve the same effect.

Thanks!

TooTallNate:
Not really. The `debug("obj: %s", util.inspect(obj, {colors: true}));` method is probably the quickest way.

PR #266, opened 2016-02-26
Expose inspect to allow custom object formatting

kevinoid:
This PR exposes the `inspect` method, which is used to format objects for the `%o` formatting directive, so that it can be called or replaced to suit the needs of users.

Some use cases for this are #185 (inspect with same coloring as debug) and #263 ("better" object formatting).  My primary use case is to log objects with a smaller depth to avoid spamming the logs from large/deep objects.  Other use cases include using the `showHidden` and `customInspect`
options of `util.inspect`.

Why not just use `util.inspect` or another custom formatting function directly?  My major reasons are:
- It takes more code since every debugged object must be formatted.
- It's not available on browsers, complicating Universal/Isomorphic JavaScript (or interfering with the inspector if polyfilled).
- It adds unnecessary formatting overhead when debugging is disabled.
- It doesn't match debug output colorization without using the undocumented `.useColors` property.

Note:  This PR currently exposes the `.inspect` property both on `debug` itself and on the instances.  It can be useful for overriding, but if you wanted to minimize the API expansion, it could be exposed as `_inspect` on `debug` or otherwise hidden.

Thanks for considering,
Kevin

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -147,6 +147,34 @@ error('now goes to stdout via console.info');
 log('still goes to stdout, but via console.info now');
 ```
 
+### Inspecting objects
+
+Objects can be included in the debug output using the `%o` formatting directive, which displays the object in the inspector on browsers and uses `util.inspect` on Node.  Objects which require special formatting can be output using `.inspect` on the debug instance.  It accepts the same options as `util.inspect` and can be overridden for each namespace:
+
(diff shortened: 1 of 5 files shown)


thebigredgeek:
Closing in lieu of https://github.com/visionmedia/debug/pull/327

PR #269, opened 2016-03-09
Make the format string goodies clearer in README

mlucool:


TooTallNate:
This is an improvement so I'll merge, but ideally we document explicitly the supported formatters.

Issue #223, opened 2015-09-13
No any output while using debug


samundrak:
I just installed and required debug in my project;
var debug = require('debug')('http');

and if i do debug('testing')
then no any output is done

but if i console.log(debug)
i see this message 
{ [Function: disabled] enabled: false, namespace: 'http' }

where i am going wrong?

dcsan:
you have to set an env var when you run your app

DEBUG="*"  node app.js

read the docs for more info on namespacing.

nikoskleidis:
This also occurred in my project too. 
I am using this:
DEBUG=app:*
in my .env file and I am using it like this in my files
import _debug from 'debug';
const debug = _debug('app:ComponentName');
debug('start component');

But I cannot see any logging in my console

dcsan:
maybe the colon throws off the regex? have you tried other ways?

nikoskleidis:
Thank you for your quick reply.
I am using this kit in my project: https://github.com/davezuko/react-redux-starter-kit#usage
I have narrowed down the problem and I realized that on the node console I see the debug as expected. The problem is on the browser's console in which I see no logging. I use google chrome.

nikoskleidis:
Figured it out! I needed to run this : localStorage.debug="app:*" on the browser in order to work properly. Thanks again for your interest!

mooyoul:
/ping @TooTallNate This is another closable issue ✅

Issue #198, opened 2015-04-19
How to get file:linenumber in debug() output?


MichaelJCole:
Is this already implemented?

Thanks!

karloluis:
Not that I know off. I believe that you can subtag your debugs though.. 

require(debug)("SUPERTAG") 
debug("subtag")("Debug notes")

But I might be hallucinating.

bucaran:
@MichaelJCole This is not possible now, but you can always `console.trace()` anywhere in your code.

@karloluis This is not possible either, unfortunately.

jwebcat:
A simple solution to this without modifying library code for Chrome, is to blackbox the debug.js script and you will still get the debugging message but you will get the line number of the js it got called from not the same line 444 from debug.js where log() is defined.

bucaran:
@jwebcat Thanks! Very useful.

jandrieu:
FWIW, (and as mentioned in issue #105 ) the blackboxing won't work for inline scripts such as created with browserify.  There's an update in the works for blackboxing in Chromium, so maybe this will get fixed soon. https://code.google.com/p/chromium/issues/detail?id=563578

Yalhu:
@jwebcat Thanks! Very helpful.

PR #273, opened 2016-04-28
Allow line numbers to match debug() call in browser consoles

fijiwebdesign:
A possible starting point to fixing file and line number mapping in the browser, #198 as well as for sourcemaps with browserify and webpack, #223 

I'm not sure what of the full formatting scope that is done in enabled() or if it can be integrated into this approach for matching line numbers. This approach has the limitation that it needs to do formatting up front when debug() is called. This is because debug() has to return console.log() directly. i.e: `return console.log.bind(console, '%c%s +%sms', 'color: ' + self.color, namespace, interval)` to preserve line numbers. 

This limits to using bind() to add formatting to console.log(). 

Currently it supports the color formatting of namespace and ms interval. The ms interval is now directly after the namespace due to limitation that we can only prepend arguments to Function.prototype.bind().

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -6,6 +6,7 @@
  */
 
 exports = module.exports = require('./debug');
+exports.is_browser = true;
 exports.log = log;
 exports.formatArgs = formatArgs;
 exports.save = save;
--- a/debug.js
+++ b/debug.js
@@ -66,6 +66,26 @@ function debug(namespace) {
   }
   disabled.enabled = false;
 
+  // override with our specific browser debug() which keeps line numbers in the console 
+  if (exports.is_browser) {
+    var interval = new Date()
+    interval.__proto__ = {
+      now: (new Date()).getTime(),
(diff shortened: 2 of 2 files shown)


neojp:
:+1: I wish this was merged. Thanks @fijiwebdesign

TooTallNate:
Very clever, but as you mention it has the limitation that formatters are disabled, which is probably a no-go, or at least a major version bump.

TooTallNate:
I think we could make something like `debug('foo %s')('bar') would, where the middle function call returns the bound console.log function with the formatter appended. Would be slower for sure, and a breaking API change.

jabaz:
This works well, but setting color manually, has no effect!

thebigredgeek:
@jabaz can you explain?

thebigredgeek:
https://github.com/visionmedia/debug/issues/370

PR #274, opened 2016-04-30
updates license date

sdaitzman:


TooTallNate:
If anything I'd hyphenate (`2014-2016`), but in any case I'm not so sure that this is necessary in the first place. My understanding for OSS license is valid from the _first_ year listed in the license to current day. Please correct me if I'm wrong.

thebigredgeek:
Yeah I agree.  @sdaitzman can you update the language to indicate a date range please?

sdaitzman:
Fixed!

PR #282, opened 2016-05-26
Prevent crash when used with react-native

vkarpov15:
Debug currently crashes when used with react-native because document is undefined (unless you explicitly set useColors): https://github.com/facebook/react-native/pull/1632

Gotta love special snowflake JS environments.


botverse:
This won't fix #303, which is a race condition problem related to the availability of documentElement.

TooTallNate:
Thanks!

vkarpov15:
@TooTallNate :beers:

PR #285, opened 2016-06-03
Remove ansi decoration

jabaz:
Adding environment variable `DEBUG_DISABLE_COLORDECOR=1` disables ansi decoration.

### DIFF
--- a/node.js
+++ b/node.js
@@ -18,7 +18,6 @@ exports.formatArgs = formatArgs;
 exports.save = save;
 exports.load = load;
 exports.useColors = useColors;
-exports.useColorDecorations = useColorDecorations;
 
 /**
  * Colors.
@@ -54,15 +53,6 @@ function useColors() {
   }
 }
 
-function useColorDecorations() {
-  var debugColorDecor = (process.env.DEBUG_COLORDECOR || '').trim().toLowerCase();
-  if (0 === debugColorDecor.length) {
-    return false
-  } else {
-    return ! /0|no|false|disabled/.test(debugColorDecor)
-  }
-}
-
 /**
  * Map %o to `util.inspect()`, since Node doesn't do that out of the box.
  */
@@ -92,13 +82,11 @@ exports.formatters.o = function(v) {
 function formatArgs() {
   var args = arguments;
   var useColors = this.useColors;
-  var useColorDecorations = this.useColorDecorations;
   var name = this.namespace;
 
   if (useColors) {
     var c = this.color;
-    var coldec = useColorDecorations ? ';1m' : 'm';
-    args[0] = '  \u001b[3' + c + coldec + name + ' '
+    args[0] = '  \u001b[3' + c + 'm' + name + ' '
       + '\u001b[0m'
       + args[0] + '\u001b[3' + c + 'm'
       + ' +' + exports.humanize(this.diff) + '\u001b[0m';
--- a/node.js
+++ b/node.js
@@ -18,6 +18,7 @@ exports.formatArgs = formatArgs;
 exports.save = save;
 exports.load = load;
 exports.useColors = useColors;
+exports.disableColorDecorations = disableColorDecorations;
 
 /**
  * Colors.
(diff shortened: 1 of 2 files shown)


stephenmathieson:
isn't this the exact same thing as `DEBUG_COLORS`?

jabaz:
No it's not, this is to be able to use colors, but disable decoration of ansi colors.
Here: https://github.com/jabaz/debug/blob/master/node.js#L103-L107
Decoration of ansi colors in this `\u001b[32;1m;` is `;1`.

Some environments doesn't support ansi color decoration.

TooTallNate:
What does the `;1` part give us? We should probably just use the most universally compatible syntax, and not have all these obscure options.

TooTallNate:
Oh I see, Bold.

DevSide:
Is there a way to detect if decoration works instead ? Otherwise no bold ?

TooTallNate:
We could perhaps take a look at the `TERM` env variable, but I doubt that will be reliable enough

thebigredgeek:
@TooTallNate any objection to this?

@jabaz can you update docs please?

TooTallNate:
I mean bold was nice, but we can't always have nice things 😄  We should make it a major version bump at least IMO

jabaz:
Yes in Windows, using Vs Code @thebigredgeek
I don't know if it can be detected.

jabaz:
It would be nice to auto detect, if supported.

thebigredgeek:
@jabaz can you update the README, I mean?

jabaz:
Yes I can, when i have some time.

thebigredgeek:
@jabaz cool!  Once done I will gladly merge this :)

TooTallNate:
> @TooTallNate any objection to this?

I'd rather not add yet anther env variable. If bold is problematic then let's just remove the bold part.

thebigredgeek:
K

jabaz:
Do you have another solution? Should i update README??

thebigredgeek:
@jabaz can you simply remove bold?

jabaz:
It's okay to me!

node.js
```diff
@@ -53,6 +54,18 @@ function useColors() {
   }
 }
 
+function disableColorDecorations() {
+  var debugColorDecor = (process.env.DEBUG_DISABLE_COLORDECOR || '').trim().toLowerCase();
+  if (0 === debugColorDecor.length) {
+    return false
+  } else {
+    return '0' !== debugColorDecor
```
DevSide: what about `return ! /0|no|false|disabled/.test(debugColorDecor)` ?
Put the regex in a constant and use it in `useColors()` too ?

node.js
```diff
@@ -53,6 +54,18 @@ function useColors() {
   }
 }
 
+function disableColorDecorations() {
+  var debugColorDecor = (process.env.DEBUG_DISABLE_COLORDECOR || '').trim().toLowerCase();
```
DevSide: disabled by default if var is not set and `process.platform === 'win32'` ?

thebigredgeek:
@jabaz any update on refactoring this PR to simply remove bold?

jabaz:
I could just set it disabled by default!

thebigredgeek:
@jabaz Explain?  Also would like to get a solution merged in by tomorrow morning

jabaz:
So that, if environment variable is not defined, it's disabled.
So it enables decoration if you add DEBUG_COLORDECOR = 1, yes, true or enabled.

thebigredgeek:
@TooTallNate thoughts on above

TooTallNate:
I don't think we need the fallback env variable (who's going to bother setting that, honestly?). Let's just remove the `1m;` and call it a day.

jabaz:
@thebigredgeek @TooTallNate So should i just remove it?

TooTallNate:
All the `DEBUG_COLORDECOR` and `useColorDecorations` stuff? Ya I think so.

TooTallNate:
It'll be a v3 feature.

jabaz:
And also remove 1m;

TooTallNate:
Yes. Just removing the bold (and thus adding more generic compatibility) would be the new feature.

jabaz:
@TooTallNate No it's done!

thebigredgeek:
@jabaz merging this into v3, which will release later this year.  Trying to avoid breaking changes for now, and realized last night while talking to @TooTallNate that a lot of people are probably parsing logs with ansi characters taken into account by the parser.

PR #297, opened 2016-07-18
Using custom logger loaded in domain.

FernandoMorais:
Adding domain capability to able folks to use custom logger (winston) loaded in a domain context.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -4,3 +4,4 @@ examples
 example
 *.sock
 dist
+.vscode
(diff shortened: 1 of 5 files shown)


DevSide:
I'm pretty sure debug doesn't want to work with a CQRS solution and an other loging library.

TooTallNate:
Why not just override the `debug.log` property? (see the [Readme](https://github.com/visionmedia/debug#stderr-vs-stdout) for an example)

thebigredgeek:
The VS file should probably also disappear

thebigredgeek:
I'm closing this for now.  Feel free to re-open if you feel that overriding `debug.log` prop (see readme) isn't a decent solution

PR #298, opened 2016-07-26
Use WorkerLocation to detect web worker

jetzhliu:
Almost all browsers support web worker also support color logger.
Also fix some environment similar to web but not support document variable(like react-native).

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -38,9 +38,10 @@ exports.colors = [
  */
 
 function useColors() {
-  // is webkit? http://stackoverflow.com/a/16459606/376773
-  return ('postMessage' in window) ||
-    ('WebkitAppearance' in document.documentElement.style) ||
+  // is webworker? https://developer.mozilla.org/en-US/docs/Web/API/WorkerLocation
+  return ('WorkerLocation' in window && location instanceof WorkerLocation) ||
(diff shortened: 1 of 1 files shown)


botverse:
Unfortunately this won't fix other similar cases as per #240

DevSide:
This PR is a first step.

@botverse, why would you consider a web worker as a specific case/environment like Electron or React Native ?

botverse:
They are examples of the same case, neither windows or document are available.

DevSide:
Web workers are in the browser scope. Two lines of code to make debug compatible them until a better solution show up looks good enough, don't you think ?

thebigredgeek:
There are conflicts, unfortunately.  Can you please update?

thebigredgeek:
@DevSide can you explain how this would work?  `window.document` doesn't exist in a web worker.  Closing this for now until we can find a more robust fix

jetzhliu:
@thebigredgeek 
This PR only make `debug` knowing that it can support color logging in a web worker environment.
It may not useful for other environment to detect whether it support color logging, but I has used this PR to work with `react-native` chrome debugging for almost half a year.

thebigredgeek:
Fixed @ https://github.com/visionmedia/debug/pull/298

PR #317, opened 2016-10-27
Added yarn integration to Makefile

thebigredgeek:


thebigredgeek:
@TooTallNate

TooTallNate:
This essentially only affects development, though I'm not sure if we should be committing/publishing the yarn.lock file

paulirish:
I asked the yarn team about this and they recommend keeping the lockfile in git.

https://yarnpkg.com/en/docs/version-control#toc-required-files sez "The following files must be checked into source control for anyone to be able "

thebigredgeek:
Yeah I also heard that we should check in the lockfile @TooTallNate @paulirish

thebigredgeek:
BTW, sorry about the branch directly in the repo.  Old habits (from work) die hard haha

TooTallNate:
I guess it doesn't hurt. Feels kind of dirty committing a "generated" file like that. Same as `npm-shrinkwrap.json` in the end I suppose.

TooTallNate:
> BTW, sorry about the branch directly in the repo.

I don't particularly mind. That's what git is for 😸

PR #321, opened 2016-11-07
Library should not contains lockfile

fengmk2:
see https://medium.com/@boennemann/avoid-yarn-for-packages-and-fully-enjoy-its-benefits-for-application-development-8bdd4deb33cf


TooTallNate:
This makes sense to me 👍

thebigredgeek:
Hmm.  Yeah good points all around.  Can you add to .gitignore and .npmignore please @fengmk2 ?

fengmk2:
@thebigredgeek Added!

thebigredgeek:
@fengmk2 thanks merging.  As a note, if you are using Yarn, use pure mode to avoid caching in lockfiles

fengmk2:
@thebigredgeek yep, I know how yarn work.

thebigredgeek:
Kk.  Just trying to be helpful :D.  I didn't know about that option until I read up more on it.  Been busy with work.  Started using Yarn to simply save time during build debugging haha

fengmk2:
Yep, yarn install is very fast after lockfile created.
We're using [npminstall](https://github.com/cnpm/npminstall) to doing the same job, and it don't need lockfile but install still very fast!

Issue #302, opened 2016-08-28
%o is not documented in Readme.md


rafis:
'%o' is not documented. Or even better would be to not to pass arguments to `util.inspect` directly, but pass them through own formatter depending on type of argument, if it is an Object or Array then format using `formatter.o`:

```
debug('myObj=', myObj); // doesn't work because '%o' is not documented
debug('myObj=%o', myObj); // works because '%o' is now documented and people knows about it
```

thebigredgeek:
@rafis would you like to create a PR for this?

thebigredgeek:
Docs should be added [here](https://github.com/visionmedia/debug/pull/322)

PR #322, opened 2016-11-07
Node: `%O` (big O) pretty-prints the object

TooTallNate:
For example:

```js var debug = require('./')('foo')

var o = {
 foo: 'bar',
 b: new Buffer(10),
 c: Math.PI
}

debug('%O', o)
```

Previously:

```
 foo { foo: 'bar', b: <Buffer 01 00 00 00 01 00 00 00 c0 82>, c:
3.141592653589793 } +0ms
```

Now:

```
 foo { foo: 'bar',
 foo   b: <Buffer 01 00 00 00 01 00 00 00 c0 82>,
 foo   c: 3.141592653589793 } +0ms
```

This is a breaking change for anybody relying on the old `%O` behavior.
Though I don't think `%O` was working previously because the formatters
regexp wasn't checking for uppercase formatters (now fixed in this patch).

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -87,6 +87,30 @@ Then, run the program to be debugged as usual.
 
   You can also exclude specific debuggers by prefixing them with a "-" character.  For example, `DEBUG=*,-connect:*` would include all debuggers except those starting with "connect:".
 
+## Formatters
+
+
+  Debug uses [printf-style](https://wikipedia.org/wiki/Printf_format_string) formatting. Below are the officially supported formatters:
+
+| Formatter | Representation |
+|-----------|----------------|
+| `%O`      | Pretty-print an Object on multiple lines. |
+| `%o`      | Pretty-print an Object all on a single line. |
+| `%s`      | String. |
+| `%d`      | Number (both integer and float). |
+| `%j`      | JSON. Replaced with the string '[Circular]' if the argument contains circular references. |
+| `%%`      | Single percent sign ('%'). This does not consume an argument. |
+
+### Custom formatters
+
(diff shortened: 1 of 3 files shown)


z-vr:
How can I print objects with many layers on one line?

TooTallNate:
> How can I print objects with many layers on one line?

Using lower-case `%o` formatter.

z-vr:
@TooTallNate it doesn't do the job though, it prints `{ func: { result: [ [Object], [Object] ] } }`

TooTallNate:
Yes I see. We need a way to configure the `depth` on a debug instance, and default to something a bit higher than `util.inspect()`'s default depth value.

TooTallNate:
^ That should probably be a separate PR though.

thebigredgeek:
@TooTallNate thinking minor semver for this?

TooTallNate:
Well considering `%O` was completely broken before, than ya minor is probably fine. The default behavior of `debug(obj)` is changing here though so anybody relying on that (for tests, for example) would be failing after an upgrade. So maybe major would be safer in that case.

thebigredgeek:
@TooTallNate kinda thinking we should expand the API for depth with this release.  Set a default value and make the api option for changing depth non-breaking.  Thoughts?

thebigredgeek:
@TooTallNate can you update the docs

thebigredgeek:
Also @TooTallNate can you add docs for [this](https://github.com/visionmedia/debug/issues/302) in the PR as well?

thebigredgeek:
@TooTallNate any update on docs man?  Also, we need to resolve conflicts on this.  I'd like to release 2.4 this week

jfoclpf:
@z-vr you may try this, it fully prints the object, but you lose colours

```
const util  = require('util');
debug(util.inspect(yourObj, {showHidden: false, depth: null}));
```

PR #325, opened 2016-11-10
fix(browser): should check whether process exists

tomtomau:
Currently, 2.3.1 breaks in the browser. We need to check whether `process` is defined.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -144,7 +144,7 @@ function load() {
   } catch(e) {}
 
   // If debug isn't set in LS, and we're in Electron, try to load $DEBUG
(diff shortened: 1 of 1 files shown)


stephenmathieson:
Nice, I just did the same: https://github.com/stephenmathieson/debug/commit/1213eb4f8a77248329526c6feac1f211938acf20

/cc @TooTallNate @thebigredgeek

TooTallNate:
Good catch. Let us know if v2.3.2 doesn't fix it for you guys.

perrin4869:
Maybe a better idea is to do something like `var process = require('process')`?

PR #328, opened 2016-11-11
Fix browser environment detection

jahed:
The Electron detection has broken Browser detection. Since browsers don't have a `process` variable, they end up include `node.js` and all the `fs`, `net`, etc. modules with it.

This should fix it, however since there aren't any tests (which would've caught this breakage), I can't be 100%.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@
  * treat as a browser.
  */
 
-if (typeof process !== 'undefined' && process.type === 'renderer') {
+if (typeof process === 'undefined' || process.type === 'renderer') {
   module.exports = require('./browser.js');
 } else {
   module.exports = require('./node.js');


TooTallNate:
What module loader are you using? It should pick up the `browser: "./browser.js"` entry in the package.json file and use that as the entry point for browsers.

jahed:
I'm using Webpack. Didn't know about the main/browser branching in npm, not sure if Webpack uses it.

TooTallNate:
It definitely does, by default at least. You might have a strange configuration...

jahed:
Hmm, I'll look into it further. The dependency on `debug` is coming from a nest of dependencies so it could be something there. Strange thing is, it's been working fine until yesterday and our config is very plain.

I'll close the PR for now.

TooTallNate:
The funny part to me is that your PR touches the latest commit in master, which hasn't even been published to npm yet. Are you pulling from master for some reason?

jahed:
The change wasn't specifically for the latest commit. More for the branching logic in the file.

The dependency on `debug` comes from [`jsonp`](https://github.com/webmodules/jsonp/commits/master) which recently had a change from a strict dependency on 2.1.3 to minor (I did a clean npm install yesterday so this change was picked up now). jsonp is relied on by [auth0.js](https://github.com/auth0/auth0.js) and so on.

Looking deeper into it, the [package.json](https://github.com/visionmedia/debug/blob/0a8e4b7e0d2d1b55ef4e7422498ca24c677ae63a/package.json) at that point also used the main/browser split. Even though this PR fixed the issue for me by directing Webpack to the browser.js via index.js, it's probably something else causing Webpack to go through the main route instead of browser in the first place.

PR #331, opened 2016-11-13
fix(browser): do not override ls debug if found

### LINKED ISSUE #330: Since 2.3.1/2.3.2 localStorage "debug" is removed every time (wrong Electron check)
After upgrading my `debug` dependency from 2.3.0 to 2.3.2 nothing works again.

My index.html runs the following code **before** running any JS library:

```html
<script>
   localStorage.setItem('debug', '*');
</script>

<!-- Load JS libs now -->
```

It has worked with any previous version of the debug library. Now it doesn't and nothing it logged.

In fact, once the the JS library (which includes the  `debug` dependency) is loaded, `localStorage.getItem('debug')` returns false.

This is critical.

levithomason:
In React apps built with Webpack `process.env.NODE_ENV` is set as a global.  This means `process.env` is a global object.  The current browser code overrides whatever is found in localStorage.  This PR returns early if a debug value is found in localStorage.

Fixes #330

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -140,15 +140,13 @@ function save(namespaces) {
 function load() {
   var r;
   try {
-    r = exports.storage.debug;
+    return exports.storage.debug;
   } catch(e) {}
 
   // If debug isn't set in LS, and we're in Electron, try to load $DEBUG
   if (typeof process !== 'undefined' && 'env' in process) {
-    r = process.env.DEBUG;
+    return process.env.DEBUG;
   }
-  
-  return r;
 }
 
 /**


TooTallNate:
LGTM /cc @thebigredgeek @ibc

Fixes #330 I think.

levithomason:
I believe so, exactly my issue.  I've closed the issue via the desription 👍

ibc:
It seems that 20c37fdc169eeb42f34d22b11362f66a6a64b19e should fix #330 issue. Hope a new NPM version 2.3.3 is released soon.

thebigredgeek:
Yeah I'll release this morning once I get to office :)

Ambroos:
@thebigredgeek Any news? 2.3.2 is broken in quite a few cases and a release would be very welcome.

PR #335, opened 2016-11-17
allow colours in workers

botverse:
1. there are no `window` or `document` globals in workers, so `useColors` throws.
2. there is still an opportunity to check if we are in webkit using the userAgent


thebigredgeek:
I am hesitant to merge something like this in without tests.  We've already had a lot of issues with workers.  Thoughts @TooTallNate ?

xuset:
Hi all! I've been having the same problem with `require('debug')` in web workers. I tried out @botverse's fix in chrome and firefox, and everything works as expected in web workers and normal contexts. Is there anything I can help with in getting this merged in?

thebigredgeek:
I am going to merge this, but it will be a minor semver.  I will release 2.4.0 tomorrow morning with this.

Issue #258, opened 2016-01-26
multiple instances on the same namespace get different colors


zacronos:
If I create multiple instances with the same namespace, they will each acquire a different color.  This is very counter-intuitive to me; if I am creating multiple loggers on the same namespace then by definition I want them to be treated as a single unit.  If I wanted them to be separate, I would if nothing else use different namespaces with a common prefix.

Example:

``` js
var debug = require('debug');
debug.enable('foo');

// shouldn't these use the same color?
var logger1 = debug('foo');
var logger2 = debug('foo');
```

I'm currently working around this in my app by wrapping `debug` with some code that caches the result by namespace, but it really seems like this functionality should be internal to `debug`.

I'm happy to submit a PR for this, if I get feedback that others agree this should be changed.

stephenmathieson:
why would you want multiple debuggers with the same name?  the entire purpose of the name is to be unique

in your example, you could do:

```js
var logger1 = debug('foo')
var logger2 = logger1

logger1('hi')
logger2('hello')
```

...but that seems pretty pointless imo

zacronos:
@stephenmathieson, of course that's a simplified example.

Agreed, the purpose of the name is to be unique, so IF someone creates 2 loggers with the same name, that means the loggers should be the same -- either 2 instances with the same color, or perhaps even get the same instance should be returned.  Neither is true right now.

As a better example, if I have 2 different modules and I wish them to use the same logger, right now I either have to tolerate them getting different colors, or else create my own way of caching/passing the logger so they can share an instance.

stephenmathieson:
> if I have 2 different modules and I wish them to use the same logger

that would defeat the purpose of naming the loggers... how would you be able to tell them apart?

zacronos:
From my perspective, the question of "how would you be able to tell them apart?" is exactly backwards.  The point is that if make 2 calls to `debug()` with the same name, I **shouldn't** be able to tell the resulting loggers apart.  I should either get the same logger back each time, or I should get 2 logger instances that have the same color/behavior (even if technically `logger1 !== logger2`).  Currently, that's not what happens.

As to explaining why I might want to do this...  it seems only tangentially relevant, but here goes.

[...]

PR #338, opened 2016-11-28
Use same color for same namespace.

lchenay:
CF #258

### DIFF
--- a/debug.js
+++ b/debug.js
@@ -42,13 +42,20 @@ var prevTime;
 
 /**
  * Select a color.
- *
+ * @param {String} namespace
  * @return {Number}
  * @api private
  */
 
-function selectColor() {
-  return exports.colors[prevColor++ % exports.colors.length];
+function selectColor(namespace) {
+  var hash = 0, i, chr, len;
+  
+  for (i in namespace) {
+    hash  = ((hash << 5) - hash) + namespace.charCodeAt(i);
+    hash |= 0; // Convert to 32bit integer
+  }
+  
+  return exports.colors[Math.abs(hash) % exports.colors.length];
 }
 
 /**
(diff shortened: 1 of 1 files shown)


TooTallNate:
Nice and simple, thanks for the PR! I'm 👍  for this feature. There's been plenty of times where the order of require calls changing ends up changing the colors of the debug instances, since they get instantiated in a different order which has always felt strange.

The only thing I'd be worried about is if the algorithm doesn't "distribute" the colors evenly enough. But I guess we'll just have to feel it out and see.

lchenay:
I agree, it will have less perfect distribution on colors over namespace.
To reduce chance of collision, as it's no more incremental, the only solution is to have more colors available.

debug.js
```diff
@@ -42,13 +42,20 @@ var prevTime;
...
  * @return {Number}
  * @api private
  */
 
-function selectColor() {
-  return exports.colors[prevColor++ % exports.colors.length];
+function selectColor(namespace) {
+  var hash = 0, i, chr, len;
+  
+  for (i in namespace) {
```
TooTallNate: http://stackoverflow.com/a/7616484/376773

Meithar:
This algorithm doesn't seem to work well. Example:
```js
  console.log({
    namespace,
    hash: Math.abs(hash),
    color: exports.colors[Math.abs(hash) % exports.colors.length],
  });
```
```js
{namespace: "rto-ext:EVENTS", hash: 1808327694, color: "lightseagreen"}
{namespace: "rto-ext:SETTINGS", hash: 1479630660, color: "lightseagreen"}
{namespace: "rto-ext:BG", hash: 33268638, color: "lightseagreen"}
```
Three different namespaces have the same color.

lchenay:
exports.colors.length == 6
So 16% of chance of collision.

@TooTallNate More color ?

Otherwise we need to have an array to store the colors stickiness. Not as complicated as this.

lchenay:
We can try something like that otherwise : https://github.com/lchenay/debug/commit/9f4404f8cd695191743e668ddd0fe4f44ca6cd34

The color will be better distributed as before, with reuse of same color for same namespace, but will use some memory to store the array of mapping.
This can lead to memory leak if a user use debug in prod with random namespace. For exemple those who use a namespace like "myapp:request:${request.expressRequestId}".

PR #342, opened 2016-12-07
Update bower.json

justmatt:
Error thrown if using `bower.json` to require files.
`.npmignore` contains `dist`.


thebigredgeek:
@justmatt we are planning on refactoring to ES6 within the next few weeks.  Once that is done, this won't be necessary as the dist will be included

PR #345, opened 2016-12-11
avoid browserifying entire process polyfill

nolanlawson:
Curently, the browser build of `debug` pulls in the entire [process polyfill](https://github.com/defunctzombie/node-process) because it's including a line checking for `process.env.DEBUG` in Electron apps.

One simple way to fix this is to do `global.process` instead of `process` which prevents Browserify/Webpack from trying to pull in the entire polyfill. I tested manually in Electron and confirmed that `global.process` is the same as `process`, so this should be a safe change.

I checked the resulting bundle size using `browserify . | uglifyjs -mc | gzip -c | wc -c` and got:

* Before: **2609 bytes**
* After: **2097 bytes** (~0.5KB saved, ~20% of total size)

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -148,8 +148,10 @@ function load() {
   } catch(e) {}
 
   // If debug isn't set in LS, and we're in Electron, try to load $DEBUG
-  if (typeof process !== 'undefined' && 'env' in process) {
-    return process.env.DEBUG;
+  // We're using global.process to avoid Browserify/Webpack including all of:
+  // https://github.com/defunctzombie/node-process
+  if (typeof global !== 'undefined' && global.process && global.process.env) {
+    return global.process.env.DEBUG;
   }
 }
 


thebigredgeek:
@nolanlawson unfortunately this breaks webpack builds in many cases, as it is common to reference `process.env` in apps compiled in webpack.  In this case, it doesn't include a process polyfill.  I think that is mainly a webpack thing.  I am hesitant to optimize around browserify in a way that might break webpack, which seems to be the "new" browserify.  Is there any way we could apply a similar fix that won't break references to "process.env"?

nolanlawson:
Hm, I wasn't sure at first what you meant, but after playing around with [Webpack DefinePlugin](https://gist.github.com/nolanlawson/4348958ada92adca6cbc4bd08cab6552) I can see that it handles `process.env.NODE_ENV` just fine, but not `global.process.env.NODE_ENV`. I'm guessing that Browserify's `envify` transform would have the same issue.

It seems to me that this is a problem not just for `debug` but for really all JS projects designed for either Webpack or Browserify. They're paying a 500 byte tax (2KB without gzip) just to support loaders/transforms that the user may or may not actually be using.

I'm going to look into solving this on my own end by using a global Browserify transform or some other tool to remove the `process` references entirely. I can understand that from your position it makes sense to just take the 500 byte tax in order to support all use cases. Thanks for the interesting discussion!

PR #346, opened 2016-12-14
add node console line number

likun7981:
add  line number console . and fix a spelling mistake


thebigredgeek:
spelling mistake was fixed in a previous commit.  Please frame new requests for features here: https://github.com/visionmedia/debug/issues/370

I think the middleware approach would likely allow you to solve such problems.

PR #356, opened 2016-12-14
Fixed a typo

diedsmiling:


mzedeler:
@thebigredgeek - can you move this out to npm soon? All our builds are broken because of this.

PR #368, opened 2016-12-14
Work around debug being loaded in preload scripts

paulcbetts:
This is a follow-on to #324, where if debug is loaded in a preload script (https://github.com/electron/electron/blob/d6e3360aeff6c50820a08af28f1097a0c25d6e3e/docs/api/web-view-tag.md#preload), it will crash in `useColors` because the DOM is not fully set-up yet (`window.document.documentElement` will be undefined, but `window.document` exists)

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -37,6 +37,13 @@ exports.colors = [
  */
 
 function useColors() {
+  // NB: In an Electron preload script, document will be defined but not fully
+  // initialized. Since we know we're in Chrome, we'll just detect this case
+  // explicitly
+  if (typeof window !== 'undefined' && 'process' in window && window.process.type === 'renderer') {
+    return true;
+  }
+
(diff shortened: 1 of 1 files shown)


thebigredgeek:
https://github.com/visionmedia/debug/issues/369 < wondering if this is the same issue?

paulcbetts:
@thebigredgeek Doesn't look like it but might be related? The callstack for us was trying to access `documentElement` of null

paulcbetts:
@thebigredgeek If you could cut a release off of this we'd be v grateful, this debug issue is causing us a lot of grief over at Slack because we use it in lots of low-level components that aren't easily patched

thebigredgeek:
@paulcbetts done :).  2.4.4 now live

paulcbetts:
Thank you so much!

thebigredgeek:
Np!  Good luck

PR #375, opened 2016-12-15
rm non-maintainted `dist/` dir

### LINKED ISSUE #372: updating dist/
I noticed `dist/` isn't being updated on each release. 

I see at least 2 options:
* Add a release script (could be extremely small 1-liner possibly) to update it, before each package.json bump and tag

* Remove it from repo, have it built/published via travis

freewil:
closes #372


README.md
```diff
@@ -142,8 +142,13 @@ debug('this is hex: %h', new Buffer('hello world'))
 ```
 
 ## Browser support
+  You can build a browser-ready script using [browserify](https://github.com/substack/node-browserify),
```
TooTallNate: We should mention that it works out of the box with the popular browser bundlers. We mention browserify but at least webpack should be mentioned as well.

freewil: I'm not familiar with webpack or anything besides browserify. Please add some info if you are able to. Otherwise, I still consider this PR an improvement as-is.

thebigredgeek:
@freewil would you mind updating the docs to mention that it works with bundlers?

PR #377, opened 2016-12-16
Fix debug calls while in a preload script inside Electron

MarshallOfSound:
In `electron` preload scripts `document.documentElement` is null, this is quite a simple fix that simply checks for a good value 👍

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -46,7 +46,9 @@ function useColors() {
 
   // is webkit? http://stackoverflow.com/a/16459606/376773
   // document is undefined in react-native: https://github.com/facebook/react-native/pull/1632
-  return (typeof document !== 'undefined' && 'WebkitAppearance' in document.documentElement.style) ||
+  // document.documentElement is null in Electron preload scripts
+  return (typeof document !== 'undefined' && document.documentElement
+          'WebkitAppearance' in document.documentElement.style) ||
     // is firebug? http://stackoverflow.com/a/398120/376773
     (typeof window !== 'undefined' && window.console && (console.firebug || (console.exception && console.table))) ||
     // is firefox >= v31?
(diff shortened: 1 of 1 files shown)


browser.js
```diff
@@ -46,7 +46,9 @@ function useColors() {
 
   // is webkit? http://stackoverflow.com/a/16459606/376773
   // document is undefined in react-native: https://github.com/facebook/react-native/pull/1632
-  return (typeof document !== 'undefined' && 'WebkitAppearance' in document.documentElement.style) ||
+  // document.documentElement is null in Electron preload scripts
+  return (typeof document !== 'undefined' && document.documentElement
+          'WebkitAppearance' in document.documentElement.style) ||
```
thebigredgeek: wondering if you meant to add an `&&` to the end of the null pointer check `document.documentElement`?  This looks like a syntax error

thebigredgeek:
Pretty sure there is a syntax error here.  I added linting to browser.js on master, but I don't think it ran on this PR.

browser.js
MarshallOfSound: Idek where that `&&` went, let's pretend this didn't happen and move on 😆

paulcbetts:
This is a dupe of #368, no?

MarshallOfSound:
@paulcbetts Hmmm, maybe I got a module before that patch or someone hasn't updated their entire dependency tree.  I'll wipe `node-modules` try again 👍

graysonlang:
Just a heads up: If you want this to not error out in the context of Electron's main process (which it did in earlier versions), you will also need to check for `typeof navigator !== 'undefined'` on lines 56 and 58.

thebigredgeek:
@MarshallOfSound conflicts :(

MarshallOfSound:
Resetting yarn cache finally pulled in a good version of this module 👍 @paulcbetts fix works fine

PR #387, opened 2016-12-20
Use typeof window.process !== 'undefined'

kribblo:
'process' in window returns true after window.process = null

Had this problem just now with a javascript game developed by a third party; they were setting window.process = null to support some other library that used that for some detection.

Maybe they are doing it wrong, but this should help if more people are also doing it wrong. :)


TooTallNate:
Thanks!

PR #388, opened 2016-12-20
Fix formatting bug

ForsakenHarmony:
Was trying to figure out why this doesn't work, turns out it was because `"%capp:signup %c%O%c +2h"` wasn't parsed properly


TooTallNate:
D'oh! Nice catch. Thanks!

PR #393, opened 2016-12-21
fix(browser): prevent ReferenceError in workers

### LINKED ISSUE #382: Web Worker compatibility
Over at jsdom, our web worker test fail if we update our dependencies. Here is the error that I got:

`ReferenceError: window is not defined` at https://github.com/visionmedia/debug/blob/17d0e0bb49b46f53bd97aa6d7b56bf9990a8cb62/browser.js#L13

This line is missing a `typeof window !== 'undefined'` check. This check _is_ present in other locations in this file, so this issue looks like a small oversight.

After #378 is resolved, you might be interested in https://www.npmjs.com/package/karma-mocha-webworker @yamikuronue

KlausTrainer:
fixes #382

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -182,6 +182,6 @@ function localstorage() {
 }
 
 /** Attach to Window*/
-if (window) {
+if (typeof window !== 'undefined') {
   window.debug = exports;
 }


snuggs:
#SALUTE!!

PR #396, opened 2016-12-25
if localStorage.debug is undefined try process.env.DEBUG

pravdomil:
Usefull for electron


thebigredgeek:
Can you add a test to make sure it tries process.env.DEBUG in this case?

pravdomil:
Well I can't, I tested on my machine and it just works.

pravdomil:
Please merge or reject this pull request.

PR #406, opened 2016-12-28
Remove DEBUG_FD

### LINKED ISSUE #280: RFC: 3.0 stdout moved off of APIs that are pending removal
It was never documented and was supposed to be internal.
Used at [node.js#L164](https://github.com/visionmedia/debug/blob/master/node.js#L164).

Refs: https://github.com/nodejs/node/pull/6749

### LINKED ISSUE #386: Proposal: remove DEBUG_FD
I implemented it a while ago, but personally have never needed to use it. Does anybody else use it?

Removing support for `DEBUG_FD` seems like it would be the simplest way to fix #280, since that's the only place where `SyncWriteStream` is used.

TooTallNate:
Now simply uses `process.stderr`. Breaking API change, for the v3 branch.

Previously used internal and undocumented Node.js APIs to support this underly used API.

Fixes #280.
Closes #386.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,6 @@ Then, run the program to be debugged as usual.
 |-----------|-------------------------------------------------|
 | `DEBUG`   | Enables/disabled specific debugging namespaces. |
 | `DEBUG_COLORS`| Whether or not to use colors in the debug output. |
-| `DEBUG_FD`| File descriptor to output debug logs to. Defaults to stderr. |
 | `DEBUG_DEPTH` | Object inspection depth. |
 | `DEBUG_SHOW_HIDDEN` | Shows hidden properties on inspected objects. |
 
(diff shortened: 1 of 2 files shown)


develar:
And how now we can change output without editing client code? What WebStorm or electron-webpack should do?

tanyaYourchuck:
@thebigredgeek @TooTallNate DEBUG_FD was removed entirely?
why not leave it whitelisted to 1 or 2 as was decided as resolution of https://github.com/visionmedia/debug/issues/410?

My project had also used DEBUG_FD to make sure debug goes to stdout in all submodules as @wallali mentioned: _Modules I depend on who also internally use debug whose code I do not control keep outputting to stderr and I cannot redirect them to stdout._

I do not want to redirect all stderr to stdout, just want all debug output to go to stdout. Is this doable without the DEBUG_FD flag?

PR #407, opened 2016-12-28
Fix implicit window access

webOS101:
The partial fix before did not address accessing `navigator`.  I whipped this up in the Github editor so you might want to reformat it.  Totally appreciate the work you guys do.  That said, this component is used by a lot of modules, in a lot of situations.  This is the 2nd week in a row where debug has broken my build process.  Expanding the test suite to test with browser and non-browser situations would be great.


webOS101:
Alternative fixes involve changing the `navigator` lines to start with:

`typeof navigator !== 'undefined'`

webOS101:
The travis error seems unrelated to the code change.

thebigredgeek:
I believe this has already been fixed.  Can you confirm @webOS101 ?

webOS101:
I believe this can be safely closed.

PR #414, opened 2017-01-16
Fixed IE8 "Expected identifier" error

vgoma:
"default" is reserved word in IE8. So it produces 'Expected identifier' error. Fixed that with quotes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+2.6.1 / 2017-01-16
+==================
+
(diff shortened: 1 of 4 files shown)


fleeco:
This makes the library compatible with IE8 - which is super useful for me.  Thanks!

vvo:
Also looking forward to this fix, any feedback @thebigredgeek?

TooTallNate:
👍  But if you could, just have it be the single line change for the quotes. We have a release process that happens in a separate commit.

PR #420, opened 2017-02-06
fix DEBUG_MAX_ARRAY_LENGTH

slavaGanzin:
```console.log(prop)```
was: ```maxArray_length```
now: ```maxArrayLength```


TooTallNate:
Nice catch! Merged.

PR #431, opened 2017-03-03
process.env.DEBUG ignored in Electron apps for all cases

paulcbetts:
This PR fixes the code introduced in #331 that breaks the Electron case, but also prioritizes LocalStorage if present


paulcbetts:
Paging @thebigredgeek to the white courtesy phone

PR #445, opened 2017-04-25
Passing settings to debug on runtime.

NicoZelaya:
We use the library on an SDK package, and we had users with issues because they were using the ENV Variable **DEBUG** for other stuff.

We needed a way to redefine the variable that debug is looking for (both for Node and Browser). I've done some extra work so we can have a settings object, instead of just defining what I needed.

Please, take a look and let me know of any changes you may need. 
Also, it would be great if this feature is included in v3! I'll read more about it, but it looks promising already :)

Best regards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,9 +129,6 @@ Then, run the program to be debugged as usual.
   [`util.inspect()`](https://nodejs.org/api/util.html#util_util_inspect_object_options)
   for the complete list.
 
-  In case the name debug cause conflicts on your project, use the settings 
-  
-
 ## Formatters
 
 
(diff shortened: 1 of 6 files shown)


thebigredgeek:
Sorry, but we can't support this and it won't be on our roadmap :(.

NicoZelaya:
Ohhh I see. Mind giving me some feedback either way? Since you own this code and you'll be able to spot issues right away :).
We need this to keep using the library so my only options are using the fork or moving to another logging utility (which I'm trying to avoid because I liked this one).

Issue #450, opened 2017-04-30
Suddenly stopped working on Node.js 0.6


dougwilson:
Up to version 2.3.3 this module worked fine in Node.js 0.6, but now it no longer does. I'm not sure if this is an issue or not, because there is nothing in the docs or even package.json that indicated what versions of Node.js are actually the supported versions.

1. Is this an issue that Node.js 0.6 stopped working?
2. What are the versions of Node.js that are supported?
3. Can the version range be at least added to engines.node in package.json?

TooTallNate:
1. We should probably fix whatever broke.
1. Virtually every node version IMO.
1. Fine by me, but perhaps kind of redundant if the goal is `"node": "*"`.

calidion:
it is most probably a problem caused by npm, where the range versions always break packages..

dougwilson:
Thanks for the clarification, @TooTallNate ! I'm happy to make a PR with a suggested fix, just wanted to check if you even cared about the version first before bothering :) I agree that adding `"node": "*"` would be redundant. It sounds to me like there is no official minimum version right now, rather leaving it up to the "market" to decide (by users making PRs with suggested fixes).

So the issue is just that `util._extend` doesn't exist in Node.js 0.6 (plus, it _is_ an underscore-prefixed function, so... _shrugs_). Anyway, without going through the complications of trying to introduce a dependency, The smallest portion of `_extend` was just inlined in PR #452

dougwilson:
Looks like the PR is merged now 🎊 ping me on the next npm publish whenever you get to it 😄

TooTallNate:
Published in 2.6.7.

PR #452, opened 2017-05-02
Inline extend function in node implementation

dougwilson:
This is a proposed solution for the Node.js 0.6 breakage I reported in #450 . This replaces the use of `util._extend` with an inlined version of the minimal implementation.

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -231,7 +231,12 @@ function createWritableStdioStream (fd) {
  */
 
 function init (debug) {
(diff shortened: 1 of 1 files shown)


TooTallNate:
Nice find @dougwilson! 👍 

I can't help but wonder what you're doing with Node v0.6 still, though 😄

dougwilson:
🤷‍♂️

PR #457, opened 2017-05-15
Fix browser detection

medikoo:
Currently in main script it is assumed that browser environment defines `process` object, that's not right:

screen shot 2017-05-15 at 10 55 00

I stumbled on it, when I tried to use debug in portable modules (ones that work on both server and client side) in a browser (via CJS bundler).

For time-being I'm relying on following [workaround](https://github.com/medikoo/onejs.org/blob/2c4a27ddf2ac04090f3055ea7e1671c9329bd472/client/lib/debug-workaround.js#L6-L8), but that's not quite beautiful.

This patch fixes it (and doesn't break proper Electron detection)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,7 +3,7 @@
  * treat as a browser.
  */
 
-if (typeof process !== 'undefined' && process.type === 'renderer') {
+if (typeof process === 'undefined' || process.type === 'renderer') {
   module.exports = require('./browser.js');
 } else {
   module.exports = require('./node.js');


TooTallNate:
I think your client-side bundler is misbehaving. [In the package.json file we have `"browser": "./src/browser.js"`](https://github.com/visionmedia/debug/blob/15850cb89debc3a2cd0b61ef7f9e298f257506cc/package.json#L42) which bundlers like browserify and webpack use by default. The `index.js` file is intended for Node.js and Electron processes to use.

medikoo:
> I think your client-side bundler is misbehaving. In the package.json file we have "browser": "./src/browser.js" 

Not all bundlers support `browser` tag, and personally I prefer to rely on cleaner (imo) approach as feature detection directly in a module (it doesn't require any custom handling from a bundler) 

I'm happy to see that you do feature detection in debug, still it doesn't seem right. Do you see any harm in change I proposed? Does it break anything?

TooTallNate:
> I'm happy to see that you do feature detection in debug, still it doesn't seem right. Do you see any harm in change I proposed? Does it break anything?

It _seems_ ok over first glance. I am skeptical though because we've learned through history of this module that tweaks to these parts of code can break edge cases. I suppose we could merge though and wait for any complaints to pop up. I still think you should use a better bundler though 😉

medikoo:
> I still think you should use a better bundler though 😉

Actually in this bundler it's more a feature than a bug :) Idea is to work with _portable_ CJS modules pure, least custom way (best if no customizations involved).

> I suppose we could merge though and wait for any complaints to pop up

I'll be grateful. The issue could appear only if some would run it in env where` typeof process === "undefined"` but still  want the `require('./node.js')` to be loaded, and that contradicts each other (as `./node.js`uses internally `process` object). So change seems 100% safe.

jedwards1211:
@TooTallNate why are you assuming everyone will be using Electron?

PR #458, opened 2017-05-16
Fix Regular Expression Denial of Service (ReDoS)

hubdotcom:
https://snyk.io/vuln/npm:ms:20170412

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "ms": "1.0.0"
+    "ms": "2.0.0"
   },
   "devDependencies": {
     "browserify": "9.0.3",


designfrontier:
would love to see this landed :-) I am a way upstream consumer and this is the one open vulnerability I've got at the moment.

thebigredgeek:
live on https://github.com/visionmedia/debug/releases/tag/2.6.7
@dougwilson

dotchev:
We also see this issue in our projects. We depend on _debug_ via a bunch of other packages. It will take some time until all of them are updated.
But I wonder if this issue affects _debug_ at al.
Looking quickly in the code, it seems `ms` is called only with a number, not with a string. So this ReDoS issue should not be relevant, right?

```js
    var ms = curr - (prevTime || curr);
    self.diff = ms;
```
https://github.com/visionmedia/debug/blob/master/src/debug.js#L73
```js
exports.humanize = require('ms');
```
https://github.com/visionmedia/debug/blob/master/src/debug.js#L14
```js
    args.push('\u001b[3' + c + 'm+' + exports.humanize(this.diff) + '\u001b[0m');
```
https://github.com/visionmedia/debug/blob/master/src/node.js#L115

PR #460, opened 2017-05-17
Make it works on React Native

ShongSu:


suark:
We just had an issue with this in our React-Native project. I manually implemented your change in this PR and it worked for me.

src/browser.js
```diff
@@ -40,7 +40,7 @@ function useColors() {
   // NB: In an Electron preload script, document will be defined but not fully
   // initialized. Since we know we're in Chrome, we'll just detect this case
   // explicitly
-  if (window && window.process && window.process.type === 'renderer') {
+  if (window && window.process && window.process.type === 'renderer' || !window.document) {
```
TooTallNate: Is this really the best feature-detection for `react-native` though? Seems fragile.

TooTallNate: Looks like we can do `navigator.product === 'ReactNative'` too.

suark: https://github.com/granduke/FE_Cab/pull/243 Does any of this help?

RangerMauve: I think that the point wasn't just to detect react-native, but to detect a lack of `window.document` which is used in the next part of the check.

RangerMauve:
I'm guessing #462 fixes this so it should be closed?

thebigredgeek:
I believe fixed https://github.com/visionmedia/debug/pull/462.  Closing

platonish:
Hi, how exactly would this work? I'm not able to see the logs on the browser running the React Native debugger.

PR #447, opened 2017-04-26
cleanup browser tests and fix null reference check on window.documentElement.style.WebkitAppearance

thebigredgeek:

PR #462, opened 2017-05-18
Check for undefined on browser globals.

marbemac:
Not all environments include these globals. For example, web workers do not have global window objects.

This seems to be a regression from this commit: https://github.com/visionmedia/debug/commit/cae07b70c968bdcadffff452dee8613522857888

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -40,20 +40,20 @@ function useColors() {
   // NB: In an Electron preload script, document will be defined but not fully
   // initialized. Since we know we're in Chrome, we'll just detect this case
   // explicitly
-  if (typeof window !== 'undefined' && window && window.process && window.process.type === 'renderer') {
+  if (typeof window !== 'undefined' && window.process && window.process.type === 'renderer') {
     return true;
   }
 
   // is webkit? http://stackoverflow.com/a/16459606/376773
   // document is undefined in react-native: https://github.com/facebook/react-native/pull/1632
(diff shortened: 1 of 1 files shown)


TooTallNate:
👍 I [mentioned](https://github.com/visionmedia/debug/pull/447#discussion_r113540420) this in the PR, but I think it got merged before it was read.

marbemac:
Ah yeah I see it there now. Yeah this little issue just bit us in the ass, because all sorts of dependencies use debug, and we're using some of them in web workers, which means they're throwing exceptions now :/.

TooTallNate:
@marbemac could you remove the second check for the tests that you've updated? We don't need a truthiness test for `window` if we're already checking for `typeof window !== 'undefined'` right before that. Then I can merge + tag.

marbemac:
I think part of the idea in that last commit was that the undefined check does not cover null. Without that second check, window.process will throw an exception if window is null.

TooTallNate:
In my experience those values will ever be `null`.

Doesn't hurt to be defensive I guess, but the point of https://github.com/visionmedia/debug/commit/cae07b70c968bdcadffff452dee8613522857888 was to clean up these checks a bit.

marbemac:
I'm happy to take it out! Shall I? Or leave it?

TooTallNate:
Let's take them out please.

marbemac:
All set!

pmeijer:
👍

pmeijer:
When will this be released?

TooTallNate:
2.6.8 published

PR #475, opened 2017-07-17
Dedicated handling of AWS Lambda environment

medikoo:
Recently I'm working with projects that stand on Serverless architecture (AWS Lambda), and found that debug doesn't play well with that.

Thing is that in AWS Lambda ( http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-logging.html ) all console.* methods are internally decorated, so then all logs are combined with meaningful info on time and invocation id), so outcome of `console.log("this is console log");` etc. ends as:

screen shot 2017-07-17 at 11 48 42

Now `debug` (as run at v2.6.8) by defaults logs as:

screen shot 2017-07-17 at 11 49 52

This patch ensures that eventual `debug` in case of being run in AWS Lambda outputs logs via `console.error` therefore comes out as following:

screen shot 2017-07-17 at 11 51 47

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -66,6 +66,8 @@ exports.inspectOpts = Object.keys(process.env).filter(function (key) {
   return obj;
 }, {});
 
+var isAWSLambdaEnv = Boolean(process.env.AWS_LAMBDA_LOG_STREAM_NAME);
+
 /**
  * Is stdout a TTY? Colored output is enabled when `true`.
  */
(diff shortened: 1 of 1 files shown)


medikoo:
> Coverage decreased (-0.7%) 

I can configure some tests that try to emulate AWS Lambda env, and run basic tests, but it may need some sophisticated handling in a module (e.g. module cache invalidation)

medikoo:
Updated for v3

medikoo:
Closing this, for following reasons:
- After #418 got in, the default output differs only in not having invocation id. which is not that relevant, as there's no concurrency within single lambda, so we can clearly deduct which invocation id applies to which debug line.
- It's too custom, which seems bad to have hard-coded in. it's best if addressed by external tweaking, which probably is doable now and if not, should be after major overhaul is done.

PR #476, opened 2017-07-17
Drop usage of `chrome.storage` (or make the storage backend pluggable)

EirikBirkeland:
The browser version assumes that chrome.storage.local uses an API which is compatible with localStorage; which is not the case. Even though I am using chrome.storage.local for my Chrome extension, I would much prefer to keep debug's variable in localStorage, as I consider chrome.storage.local the 'private space' of my extension.

This change obviates the need to support multiple storage types. But if storage type is important, how about supporting a custom storage facility, including chrome.storage.sync? I.e. the user would provide an object that follows the conventions. Just a thought - I certainly don't see the need at this point though.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -10,10 +10,7 @@ exports.formatArgs = formatArgs;
 exports.save = save;
 exports.load = load;
 exports.useColors = useColors;
-exports.storage = 'undefined' != typeof chrome
-               && 'undefined' != typeof chrome.storage
-                  ? chrome.storage.local
-                  : localstorage();
+exports.storage = localstorage();
 
 /**
  * Colors.


thebigredgeek:
I think the idea of using a "default" storage type and providing a way to "plug in" your own is a worthy idea.  For now, though, this is a breaking change for existing extension support as the storage wouldn't be extensions-specific.

EirikBirkeland:
I see your concerns.
But actually it's already broken, so removing chrome.storage.local would actually be a great thing:
chrome.storage.* uses an async API not compatible with the arraylike access afforded by localStorage. 

Therefore, removing it and keeping localStorage / env variable as the built-in solution would be the best option. This would not be a breaking change - it's already broken for those developing extensions such as myself.

API reference:
https://developer.chrome.com/extensions/storage

hdodov:
Please merge, you guys. Using `chrome.storage.local` incorrectly makes the module completely useless when you have the `storage` permission enabled in your extension.

EirikBirkeland:
Waiting half a year now man :D Actually, I have no immediate need for this now. But bump.

Thanks for `debug` - we're using it, and since we aren't building Chrome extensions right now, it works great.

Qix-:
Thanks @EirikBirkeland!

lidel:
When will there be a release with this fix?

EirikBirkeland:
I'd fork or pin to Github commit :)

PR #481, opened 2017-08-08
Support 256 colors

Qix-:
Only having a select few colors gets old and can make certain namespace prefixes harder to read (especially when they have similar lengths and are the same color).

This PR enables 256 colors when both of the following are true:

- `supports-color` is available
- `require('supports-color').level >= 2`

Otherwise, the color level falls back to the current selection of basic colors.

---

To generate the list of colors, I selected all ANSI 256 colors that satisfy the following requirements:

- HSV saturation of at least 70
- HSV value of at least 80
- The `min` of the contrast against both white and black is no less than 1.5

You can use the following script to experiment:

```javascript
const Color = require('color');

const cols = [];
for (let i = 0; i < 256; i++) {
	const c = Color.ansi256(i);
	
	if (c.saturationv() < 70) {
		continue;
	}

	if (c.value() < 80) {
		continue;
	}

	const minContrast = Math.min(c.contrast(new Color('white')), c.contrast(new Color('black')));
	if (minContrast < 1.5) {
		continue;
	}

	console.error(`\x1b[48;5;${i}m    \x1b[m`, i, '\t', c.saturationv(), c.value(), c.contrast(new Color('white')), c.contrast(new Color('black')));
	cols.push(i);
}

console.log(cols.join(', '));
```

---

How I personally will be using this modification is to add `supports-color` to my dev dependencies since it's not required in production. Great for local development with no overhead in prod.


TooTallNate:
This is awesome! A few comments:

1. Needs Readme docs about the optional `supports-color` dep (it's lame because `peerDependencies` would output a warning upon `npm i` which we wouldn't want in this case, so meh we can't use that feature).
1. The screenshots in the Readme are woefully outdated. This is especially true if/when we merge #418. I'm not asking you to do it, but just something I want to keep in mind.
1. It would be cool to generate the equivalent CSS colors for the browser implementation. We can basically _always_ output 256 color mode since we can use arbitrary CSS colors in the browser.

TooTallNate:
For comparison, before:

screen shot 2017-08-08 at 9 47 45 am

After `Qix-:256-colors`:

screen shot 2017-08-08 at 9 46 53 am

ch-andrewrhyne:
Love it.  Let's get the build passing?

thebigredgeek:
Love it!  Let's get the build passing?

Qix-:
@thebigredgeek https://travis-ci.org/visionmedia/debug/jobs/262103770 <- re-run that test :) You have coveralls as a part of the pass/fail steps, and coveralls likes to fail from time to time, randomly. If you re-run it, it should pass.

Qix-:
@TooTallNate I updated the colors to be color bounded so they look better on white backgrounds.

PR #484, opened 2017-08-08
[PROPOSAL] Release v3

TooTallNate:
This is what I think we should merge for a v3.0.0 release. It's not everything we've been talking about for v3, but I think it has enough breaking changes at this point that we should try to get them out there and normalize before any more big changes.

Breaking changes:

 * remove `DEBUG_FD`
 * uses `Date.toISOString()` for non-tty timestamps
 * `enable()` works on already existing `debug` instances dynamically
 * the millisecond diff timer is now per-namespace, rather than global

New features:

 * New `destroy()` function on debug instances for when you're dynamically creating an instance that should be garbage collected later. For example, if you're making one debug instance per HTTP request, then you would call the `debug.destroy()` function once the HTTP request has completed
 * "256 colors" (76 more specifically, so that they look good on both black and white backgrounds) support for namespace colors, makes for more variance than 6 colors
 * Refactored the README.md file with updated screenshots and more relevant information
 * Separated the `example` dir into `examples/node` and `examples/browser` for better organization


HarrySarson:
It may now be too late, but is a `destroy()` method necessary? I couldn't find any usage of `debug.instances` in the source and so was wondering if there was a reason that all instances were storing in the array.
If the `debug.instances` array was removed then the destroy method would become redundant and the garbage collector would remove debug instances when they are no longer needed.

TooTallNate:
@HarrySarson `debug.instances` is uses by the `enable()` function. In most "normal" use cases of debug you will not need to use it (i.e. when creating a single debug instance at the top of the module), but if you're creating instances dynamically (inside every HTTP request or inside of a class constructor, for example) then you'll need the `destroy()` function to prevent memory leaks.

HarrySarson:
I feel that users should not have to explicitly `destroy()` debug instances it should be enough make the debug instance inaccessible and let the garbage collector clean up.

Is having to manually destroy instances putting too much responsibility on the user?

TooTallNate:
Believe me, I wanted to make dynamically enabling and disabling existing debug instance possible without having to keep an array of them, but it's just not possible. Closest I was able to get is with WeakMap, but the problem there is that you can't iterate over a weak map like you can with a regular map so that idea didn't work.

But like I said, the use-case for when you actually need to call `destroy()` is an edge case: you only need to call it when you're creating instances dynamically, for example in a class constructor or during the lifecycle of an HTTP requests.

For the most common use-case, creating a single debug instance at the top of the module, you will not need to call destroy since you want debug to work for the duration of the Node process.

HarrySarson:
Ah that's fair :) I believe you. Thanks for the explanation

PR #485, opened 2017-08-09
Add timed sections

Qix-:
This adds timed sections, similar to `console.time()` and `console.timeEnd()`.

I understand completely if this isn't something you guys want added. However, it should be a non-breaking change.

If it _is_ something you're interested in, I'll add docs to the readme :)

This works both in the browser and in Node.

---

Usage:

```javascript
const debug = require('debug')('foo:bar');

const sec = debug.begin('section 1');

sec.mark('this is a marker');
sec.mark('it, too, shows the delta from the beginning of the section');

sec.end('finally, we can end the section');
// or simply
sec.end();
```
screen shot 2017-08-09 at 2 18 25 am

As you'd expect, `.mark()` is entirely optional, and the arguments to `.mark()` and `.end()` are optional as well.

Arguments passed to `.mark()` and `.end()` are separated from the `.begin()` text with `::`, and all formatting parameters are preserved as you'd expect - even in the browser:

screen shot 2017-08-09 at 2 19 27 am

### DIFF
--- a/src/debug.js
+++ b/src/debug.js
@@ -21,7 +21,6 @@ exports.instances = [];
 /**
  * The currently active debug mode names, and names to skip.
  */
-
 exports.names = [];
 exports.skips = [];
 
@@ -63,7 +62,7 @@ function createDebug(namespace) {
 
   var prevTime;
 
-  function debug() {
+  function debugHandle(rawArgs, section) {
     // disabled?
     if (!debug.enabled) return;
 
@@ -78,9 +77,9 @@ function createDebug(namespace) {
     prevTime = curr;
 
     // turn the `arguments` into a proper Array
-    var args = new Array(arguments.length);
+    var args = new Array(rawArgs.length);
     for (var i = 0; i < args.length; i++) {
-      args[i] = arguments[i];
+      args[i] = rawArgs[i];
     }
 
     args[0] = exports.coerce(args[0]);
@@ -109,10 +108,14 @@ function createDebug(namespace) {
     });
 
     // apply env-specific formatting (colors, etc.)
-    exports.formatArgs.call(self, args);
+    exports.formatArgs.call(self, args, section);
 
     var logFn = debug.log || exports.log || console.log.bind(console);
     logFn.apply(self, args);
+  }
+
+  function debug() {
+    debugHandle(arguments);
   }
 
   debug.namespace = namespace;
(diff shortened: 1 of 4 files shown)


Qix-:
A part of me wants to right-pad the `[end]` and `[mark]` labels so the debug message left-aligns with all section-related text. I think it'd look cleaner, and it's a trivial change. Thoughts?

Qix-:
An alternative/additional syntax could be:

```javascript
debug.time('some procedure', () => {
    // ... some lengthy operation
});
```

Qix-:
https://travis-ci.org/visionmedia/debug/jobs/262639220 needs to be re-run. :)

Qix-:
Latest commits do the following:

---

```javascript
const sec = debug.begin('section');
sec.mark(); // OK
sec.mark(); // OK
sec.end();  // OK
sec.mark(); // no output
sec.end();  // no output
```

May be reasonable to throw instead of swallow the output.

---

```javascript
debug.time('some function', sec => {
    // work...
    sec.mark('some mark');
    // more work...
    sec.end('some result');
});
```

TooTallNate:
@Qix- Very interesting! I did [something similar](https://github.com/TooTallNate/debug-time/blob/master/index.js) (though a lot more basic) using a custom formatter. That said, my initial question is if this is something that could live as a separate module like my `debug-time` thing.

Qix-:
The mechanics rely heavily on the formatter and the indirection of the main debug function to be efficient. I don't know if you could do that without a complete fork.

src/debug.js
```diff
@@ -31,7 +30,15 @@ exports.skips = [];
  * Valid key names are a single, lower or upper-case letter, i.e. "n" and "N".
  */
 
-exports.formatters = {};
+exports.formatters = {
+  s: String,
+  i: function (v) {
```
TooTallNate: `parseInt(v, 10)`?

Qix-:
```javascript
const o = { valueOf: () => 1234 };

parseInt(o, 10); //-> NaN
Number(o); //-> 1234
```

TooTallNate:
Let's get this merged @Qix-.

Needs Readme docs and there appears to be a bug where the marked sections fall back to a regular debug() call when stdout is not a TTY (server logs, etc.) that needs to be fixed first as well.

DiegoRBaquero:
This is a really nice feature. @Qix-

rmwxiong:
Any plans to get this in @Qix- ?

Issue #453, opened 2017-05-10
Optionally disable timestamps


silverbucket:
When a node process is managed by systemd, systemd already provides timestamps in addition to the automatic prepending of timestamps by debug when not in a TTY. Therefore each line has double-timestamps.

```
Mar 19 16:29:19 dev sockethub[8187]: Sun, 19 Mar 2017 16:29:19 GMT express:router:layer new '/complete'
Mar 19 16:29:19 dev sockethub[8187]: Sun, 19 Mar 2017 16:29:19 GMT express:router:route get '/complete'
Mar 19 16:29:19 dev sockethub[8187]: Sun, 19 Mar 2017 16:29:19 GMT express:router:layer new '/'
Mar 19 16:29:19 dev sockethub[8187]: Sun, 19 Mar 2017 16:29:19 GMT express:router:route new '/delayed'
Mar 19 16:29:19 dev sockethub[8187]: Sun, 19 Mar 2017 16:29:19 GMT express:router:layer new '/delayed'
Mar 19 16:29:19 dev sockethub[8187]: Sun, 19 Mar 2017 16:29:19 GMT express:router:route get '/delayed'
Mar 19 16:29:19 dev sockethub[8187]: Sun, 19 Mar 2017 16:29:19 GMT express:router:layer new '/'
```

`DEBUG_NOTIMESTAMPS=1` or something similar would be a great feature to have.

EirikBirkeland:
Timestamps show up in PM2 also. I am interested in toggleable, namespaced logging, not timestamps.

amejiarosario:
This is an issue in `Heroku` as well. I have double dates in the logs 👎 
cc: @TooTallNate @EirikBirkeland

EirikBirkeland:
I forked and changed the code, works for me :) Thanks though.

Qix-:
Closing in favor of #582. Please move discussion there.

PR #486, opened 2017-08-12
Hide tty date

amejiarosario:
This option is very needed when running Node on systems that already provides time stamps such as Heroku, PM2, systemd...

Solves #453


ptusch:
*Dump* Would really love this feature as `debug` poisons our logs for now :s

amejiarosario:
Thanks! As mentioned before, this env var is very needed for avoiding the double timestamps in Heroku, PM2, Systemd...

TooTallNate:
Thank you! Merged in 56a3853b95990a22079d646601aa01e93eceb1c7.

PR #488, opened 2017-08-19
Ability to suppress timestamps when useColors is false - maintains backwards compatibility

rynmsh:
To provide the option to disable printing of timestamps when userColours is false.  Will maintain the default behaviour if its not set for backwards compatibility. 

Reasons for this change have been mentioned in some comments already -  the need to disable the double logging of timestamps in systems that already print timestamps. 

https://travis-ci.org/rynmsh/debug


Qix-:
Closing in favor of #582. Please comment there.

PR #500, opened 2017-09-03
Adds package lock file.

HarrySarson:
npm advises that `package-lock.json` be checked into version control.

See https://docs.npmjs.com/files/package-locks#using-locked-packages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7022 @@
+{
+  "name": "debug",
+  "version": "3.0.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "acorn": {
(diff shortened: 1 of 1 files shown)


TooTallNate:
Thanks for the PR! 💜

Frankly I don't agree that `package-lock.json` should be committed to version control in the context of modules. For end-applications it makes sense but not for modules, where it just gets ignored upon `npm install` anyways.

HarrySarson:
Fair. :)

HarrySarson:
In that case should package-lock.json be added to gitignore? Otherwise it hangs around as an versioned file.

TooTallNate:
Ya good call e7e568a24736486721882282eb21beb31c741647

PR #504, opened 2017-09-21
fix: remove ReDoS regexp

### LINKED ISSUE #501: Vulnerable Regular Expression
The following regular expression used in the "o" formatter is vulnerable to ReDoS:
```js
/\s*\n\s*/
```
The slowdown is moderately low: for 50.000 characters around 2 seconds matching time. However, I would still suggest one of the following: 
- remove the regex,
- anchor the regex,
- limit the number of characters that can be matched by the repetition,
- limit the input size.

If needed, I can provide an actual example showing the slowdown.

zhuangya:
- split the line by '\n',
- and trim each line(do we really need to trim each line?).
- then join each line by ' '

i think this is the same as the original RegExp, correct me if wrong.

fix #501

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -83,7 +83,9 @@ function useColors() {
 exports.formatters.o = function(v) {
   this.inspectOpts.colors = this.useColors;
   return util.inspect(v, this.inspectOpts)
-    .replace(/\s*\n\s*/g, ' ');
+    .split('\n').map(function(str) {
+      return str.trim()
+    }).join(' ');
 };
 
 /**


TooTallNate:
Thank you!

dougwilson:
Can this be backported to the 2.x series?

TooTallNate:
@dougwilson Sure thing: https://github.com/visionmedia/debug/releases/tag/2.6.9

Curious what's holding you back from upgrading though 😅

worakanya1985:
redos

jcmarcfloress:
Thank  you

PR #506, opened 2017-09-22
fix: treat * as wildcard for `enabled()`

### LINKED ISSUE #495: Can't Disable debug
https://runkit.com/ilyaigpetrov/59a6ef385d75ca0011ffb2c3
```js
const debug = require('debug')
console.log('ENABLED_1?', debug.enabled('*'));
debug.enable('*');
debug.disable('*');
console.log('ENABLED_2?', debug.enabled('*'));
```
Output:
```js
ENABLED_1?
true
ENABLED_2?
true
```

zhuangya:
fix #495

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -157,8 +157,8 @@ module.exports = function setup(env) {
   function enable(namespaces) {
     createDebug.save(namespaces);
 
-    createDebug.names = [];
-    createDebug.skips = [];
+    createDebug.names = createDebug.names || [];
+    createDebug.skips = createDebug.skips || [];
 
     var i;
     var split = (typeof namespaces === 'string' ? namespaces : '').split(/[\s,]+/);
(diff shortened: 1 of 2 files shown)


natew:
+1, this is sort of a big feature for a logging library :)

thebigredgeek:
Docs need to be updated and conflicts need to be resolved :)

zhuangya:
> Docs need to be updated and conflicts need to be resolved :)

done =)

src/common.js
```diff
@@ -157,8 +157,8 @@ module.exports = function setup(env) {
   function enable(namespaces) {
     createDebug.save(namespaces);
 
-    createDebug.names = [];
-    createDebug.skips = [];
+    createDebug.names = createDebug.names || [];
```
TooTallNate: Why are these two necessary? Won't `createDebug.names` always be present?

zhuangya: namespace support need to keep old names, if `createDebug.names = []`, would break here: https://github.com/visionmedia/debug/pull/506/files/439e358fd48adb95d373e525d1dc1cffac4d20b0#diff-ec54818fc97b81da761d4b169c52c0ecR80

Issue #483, opened 2017-08-08
debug with chrome dev tool  issue


shivshankar3578:
debug @latest
i hit with follow with node 

> DEBUG=picnicapp:* node --inspect-brk  ./bin/www

```
const debug = require('debug');
	log = require('debug')('picnicapp:stdout')
	debug.log = console.log.bind(console);
log("test custom console");
```

console  look like
>   [33;1mpicnicapp:stdout [0mtest custom console [33m+0ms[0m

anyone can help me ?

ibc:
This seems to be an encoding error not related to `debug` module itself. Try the same by just using `console.log()`.

shivshankar3578:
although It's working linux console but not in chrome console. as you mention in readme

> Web Browser
> 
> Colors are also enabled on "Web Inspectors" that understand the %c formatting option. These are WebKit web inspectors, Firefox (since version 31) and the Firebug plugin for Firefox (any version).

and my chrome support 

> console.log("%c this is blue text", "color:blue")

then this should work. i think chrome dev tool relativly new. improve "debug" to work with this powerful tool  
thanks

TooTallNate:
Ya I see the problem when debugging a Node app with `--inspect-brk`. The problem is that I can't seem to figure out a way to change the arguments passed to `console.log()` depending on if it's the terminal or the web inspector.

If you can come up with any clever ways to handle that then I'm all ears.

Qix-:
Wrote a hack for you. Should work in any version of Node with adequate API support, failing back silently to the current functionality.

```javascript
function formatInspector(...args) {
	// TODO format the message for the web inspector
	return `INSPECTOR: ${JSON.stringify(args)}`;
}

function formatNodeConsole(...args) {
	// TODO format the message for the Node.js console
	return `CONSOLE: ${require('util').inspect(args, {colors: true})}`;
}

function hookConsole(inspectFormatter, nodeFormatter) {
	const defaultConsole = console;
	const defaultStdout = console._stdout;
	const defaultStderr = console._stderr;

	try {
		const configBinding = process.binding('config');
		if (!configBinding.debugOptions.inspectorEnabled) {
			return console.log;
		}
	
		const inspectorConsole = global.console;
		
		const vanillaConsole = new (require('console')).Console(console._stdout, console._stderr);
	
		if (!(delete global.console)) {
			return console.log;
		}
	
		global.console = vanillaConsole;

		const nullStream = new (require('stream').Writable)();
[...]

PR #507, opened 2017-09-22
refactor to make the common code be a setup function

### LINKED ISSUE #468: Cannot use debug with Rollup
There're a troubles both using `debug` directly (via `import debug from 'debug'`) and as a dependency to another package, with Rollup bundler.

1. `import debug from 'debug'`:
> Non-existent export 'default' is imported from node_modules/debug/src/browser.js

I believe this is because of [that dynamic stuff](https://github.com/visionmedia/debug/blob/52e1f21284322f167839e5d3a60f635c8b2dc842/src/browser.js#L7-L11).

2. `import SocketIo from 'socket.io-client'` which has `debug` as a dependency:
> 'default' is not exported by 'node_modules/socket.io-client/node_modules/debug/src/browser.js'

As far as I understood, the main idea behind `browser.js` is to import kinda «base» debug object and upgrade it with browser-specific methods' implementations. I believe this can be rewritten in more static and bundler-friendly way. Like requiring base, propagate it with methods and then export as a default in different statements which would simplify the analysis by Rollup and other-like.

I'm not an active `debug` user, but `debug` is very popular tool, so many of my packages' of choice using it as a dependency. So I stumble upon it when bundling, from time to time. Leveraging any workarounds is diffcult, since I do not have precision control over dependencies of my dependencies.

TooTallNate:
I believe this will fix #468.

It also enables us to be able to create both a Node.js instance and
web browser instance for when `--inspect` is used in Node.js (#483).


StreetStrider:
@TooTallNate both `6ab9525c9841b656d996e521cf86192d5647483a` and current npm latest works well with Rollup now. 🤔

src/common.js
```diff
@@ -0,0 +1,235 @@
...
+   * @param {String} namespace
+   * @return {Number}
+   * @api private
+   */
+
+  function selectColor(namespace) {
+    var hash = 0, i;
+
+    for (i in namespace) {
+      hash  = ((hash << 5) - hash) + namespace.charCodeAt(i);
```
thebigredgeek: Woah haha!

homerjam:
Hey! Is this going to get released soon? 🙏

bitflower:
Yeah, we need this also in StencilJS ! ;-)

dmnsgn:
Hi @TooTallNate , could we get this published to npm? It seems that 3.1.0 doesn't have this and rollup isn't working without this fix. Thanks!

Issue #425, opened 2017-02-22
`debug.enable()` flushes enabled namespaces


dbo:
Calling `enable` is disruptive to any enabled namespaces, which has changed in the recent minor version, caused by https://github.com/visionmedia/debug/pull/409

```bash
$ DEBUG=foo node -e 'var dbg = require("debug"); dbg.enable("bar"); console.log(dbg.enabled("foo"))'
false
```

geonanorch:
The new implementation allows full control (i.e. override of the original `DEBUG` settings), which seems desireable (even more in the context of changing this dynamically later on, see issue #433 ).
I submit that the main problem here is that public method `enable()` is not documented.
Actually `debug` lacks an API reference document &ndash; does obviously not prevent it from being appreciated and widely used, but makes integration with other tools more difficult...

Qix-:
I disagree with @geonanorch slightly - I feel like `.enable()` should be able to take `-*` and `.disable()` should be able to take `*` in order to clear the enabled namespaces.

In fact, the default behavior if no namespaces are passed (e.g. literally `debug.enable()` and `debug.disable()` alike) should be to implicitly pass `*`, which would enable/disable all namespaces, respectively.

geonanorch:
@Qix- I don't think that we disagree, your proposal for a reset makes plenty of sense!
Personally I prefer the 'no parameter provided' approach to a wildcard, i.e. `.enable()` rather than `.enable(*)`, but that is cosmetic really.

Qix-:
Well realistically both should be supported, I think.

mblarsen:
It seems this issue is a wont-fix? @Qix- @geonanorch ?

In other words: the bug is a feature :)

The implicit and explicit `*` and `-*` seems to be a different feature/issue.

Qix-:
@mblarsen Definitely not a wontfix, it's just a breaking change and requires a lot of work.

mblarsen:
Related #451

mblarsen:
`.disable()` is nice for convenience and code readability though.

Qix-:
But it's a redundant API call. Perhaps a better name for `.enable()`, then. Having multiple ways of doing the exact same thing is poor API design.

360disrupt:
> and that calls to `.enable()` and `.disable()` should instead modify the existing namespaces.

[...]

PR #517, opened 2017-10-13
docs: document enable()

abenhamdine:
Explain usage for undocumented `enable` method.
It would help to avoid pitfall as in https://github.com/visionmedia/debug/issues/425

Don't hesitate to reword if need be.


TooTallNate:
Thanks!

PR #521, opened 2017-11-08
supportsColor.level => supportsColor.stdout.level

cw881014:
you wrote wrong

```
console.log(supportsColor.level)
undefined

console.log(supportsColor.stdout.level >= 2)
true
```

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -24,7 +24,7 @@ exports.colors = [ 6, 2, 3, 4, 5, 1 ];
 
 try {
   var supportsColor = require('supports-color');
-  if (supportsColor && supportsColor.level >= 2) {
+  if (supportsColor && supportsColor.stdout.level >= 2) {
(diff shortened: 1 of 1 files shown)


TooTallNate:
This is because of `supports-color@5` [breaking API change](https://github.com/chalk/supports-color/pull/64) (/cc @Qix-).

It's probably easy enough to check both APIs for now in `debug`?

Qix-:
Yeah so use `(supportsColor.stdout || supportsColor).level` and then probably use stderr because I believe we use it instead of stdout (can you confirm @TooTallNate?)

TooTallNate:
Yes, `supportsColor.stderr` please.

realityking:
This can be closed, it was fixed with 285dfe10a5c06d4a86176b54bef2d7591eedaf40

Issue #399, opened 2016-12-26
[RFC] child debug instances


thebigredgeek:
discuss!

```javascript
const log1 = debug('foo');
log1.use((data, next) => ... ); //apply some new middleware
const log2 = log.child('bar');
// uses middleware of ancestors unless overridden on this instance
log2('hello world'); // => foo:bar hello world
```

DiegoRBaquero:
This is a really nice idea

Qix-:
Aside from the middleware aspect (#398), this was actually added a while back :)

PR #524, opened 2017-11-23
instance extends feature

kwolfy:
This adds the ability to dynamically create "child-loggers"
This is just add-on, not breaking-change

Example

```js

const log = debug('auth');
const logSign = log.extend('sign');
const logLogin = log.extend('login');

log('hello'); // auth hello
logSign('hello'); //auth:sign hello
logLogin('hello'); //auth:login hello
```

It's useful when there is one class with large functions that need to be divided into different namespace but at the same time be able to easily log all of them

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -148,11 +148,7 @@ module.exports = function setup(env) {
   }
 
   function extend (namespace, delimiter) {
-    if(typeof delimiter === 'undefined') {
-      delimiter = ':';
-    }
-
-    return createDebug(this.namespace + delimiter + namespace);
+    return createDebug(this.namespace + (delimiter || ':') + namespace);
   }
 
   /**
--- a/src/common.js
+++ b/src/common.js
@@ -148,7 +148,7 @@ module.exports = function setup(env) {
   }
 
   function extend (namespace, delimiter) {
-    return createDebug(this.namespace + (delimiter || ':') + namespace);
+    return createDebug(this.namespace + (typeof delimiter !== 'undefined' ? delimiter : ':') + namespace);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


Qix-:
Closes #399.

src/common.js
```diff
@@ -146,6 +147,14 @@ module.exports = function setup(env) {
     }
   }
 
+  function extend (namespace, delimiter) {
+    if(typeof delimiter === 'undefined') {
+      delimiter = ':';
+    }
+
+    return createDebug(this.namespace + delimiter + namespace);
```
Qix-: You can remove the if check and just change the `+ delimiter +` to `+ (delimiter || ':') +`

Qix-:
Your tests appear to be failing. Could you [check on those](https://travis-ci.org/visionmedia/debug/jobs/395332542)

kwolfy:
@Qix- I fix it

thebigredgeek:
This is clever.  I love it.  @TooTallNate ?

DhavalW:
I've implemented child instances similarly, in [debug-pest](https://github.com/DhavalW/debug-pest/graphs/traffic), which is a light wrapper over debug.
Along with log levels (log, warn, error), 
& the ability to swap debug in & out without breaking anything.

I did break convention a little bit - to include log levels as the topmost namespace. 
ie `logs:octopus:endpoint, errors:octopus:endpoint ` 

so that logs & warnings may be shut off while errors are allowed to show, for all modules in a single shot. ie - `errors:*, -logs:*, -warnings:*`

Does that make sense ? Could there be any issues with it ?
Also, happy to contribute that code here, if you guys like it.

Qix-:
Landing this, though please see https://github.com/visionmedia/debug/issues/598#issuecomment-416037696.

kwolfy:
Thank you!

MalyshevDmitry:
Thanks a lot!

PR #525, opened 2017-11-27
Export getDate so can be overwritten

reggi:


Qix-:
I'm :-1: on this - we need a better, more standardized way to change the date format. See #582 for some of this.

Thoughts @TooTallNate and @thebigredgeek?

PR #528, opened 2017-12-03
Provide a general millisecond diff

DiegoRBaquero:
This is more useful than per namespace ms diff

This way you can detect hotspots, blocking code, etc.


Qix-:
> This is way more useful

I personally disagree, but this is a subjective change. Please move this discussion to #582 as I think a delimiter for a namespaced diff and a non-namespaced diff would fix your problem.

PR #542, opened 2018-01-17
fix memory leaks caused by acitive debug instances when a debug target is not enabled

BUPT-HJM:
Recently, I've experienced serious memory leaks problem when I use [nsqjs](https://github.com/dudleycarr/nsqjs). I debug it and find the root of memory leak problem in the end.

[nsqjs](https://github.com/dudleycarr/nsqjs) uses debug@3.0.1, and it doesn't destroy active debug instance when the debug target is not enabled. `debug` module will keep the debug instances active when the debug target is not enabled. I think it is not a good idea. In my opinion, it should't have side effects when it is disabled.

I comment out `createDebug.instances.push(debug);`(https://github.com/visionmedia/debug/blob/master/src/common.js#L134)  and then I get the image as follows.

The memory leak is gone.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -131,8 +131,10 @@ module.exports = function setup(env) {
       createDebug.init(debug);
     }
 
-    createDebug.instances.push(debug);
-
+    if (!debug.enabled) {
+      createDebug.instances.push(debug);
+    }
+   
     return debug;
   }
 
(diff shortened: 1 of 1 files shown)


BUPT-HJM:
Thank you for your reply. I will close the PR.
In my option, the document should mention it in particular so that other people will not meet the same memory leaks problem as me.

TooTallNate:
You are very correct about the documentation. I'll get on that. Patch welcome if you beat me to it.

PR #543, opened 2018-01-23
introduce `debug.id` helper method

indutny:
When debugging the applications that work with pools of streams
or sockets, the debug stream often comes out interleaved making it
impossible to distinguish between streams. It is a common knowledge
that unique `id`s could be assigned to each stream and be used in
logging to solve this. However, this solution is rather ad-hoc and
usually applied only during manual debugging.

This commit introduces standard function `debug.id()` that
increments the internal counter only when debugging is enabled, and
otherwise returns `undefined`.

---

This have bugged me any times, but realization that it could be done in one place and for all came just now.

As a more questionable alternative, I think it could work differently too:

```js
const debug = debug('app');

class Instance {
  constructor() {
    this.debug = debug.id('id=%d msg: %s');
  }

  method() {
    this.debug('message params=%j', {});
    // "app id=123 msg: message params={}"
  }

  workWithAnotherInstance(other) {
    this.debug('work with=%d', other.debug.id);
    // "app id=123 msg: work with=456"
  }
}
```

 What do you think?

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -65,6 +65,7 @@ module.exports = function setup(env) {
 
   function createDebug(namespace) {
     var prevTime;
+    var lastId = 0;
 
     function debug() {
       // disabled?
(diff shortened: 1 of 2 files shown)


indutny:
Going to open a second PR for alternative.

Qix-:
Hello @indutny :)

What does the ID function solve? I mean, what would consume it?

Qix-:
https://github.com/visionmedia/debug/pull/544#issuecomment-448419084

PR #544, opened 2018-01-23
introduce `debug.unique()`

indutny:
When debugging the applications that work with pools of streams
or sockets, the debug stream often comes out interleaved making it
impossible to distinguish between streams. It is a common knowledge
that unique ids could be assigned to each stream and be used in
logging to solve this. However, this solution is rather ad-hoc and
usually applied only during manual debugging.

Introduce `debug.unique([ format ])` method that returns function with
the same signature as `debug`. This function however prepends either
`'%d '` or `format + ' '` (depending on the presence of `format`) to the
first argument of original `debug` function, and supplies unique
integer as a second argument, shifting the rest to the right.

Example:

```
const debug = require('debug')('ns');

class Instance {
  constructor() {
    this.debug = debug.unique('id=%d ');
  }

  method() {
    this.debug('method data=%j', {});
    // "id=123 method data={}"
  }

  attach(other) {
    this.debug('attach to=%d', other.debug.id());
    // "id=123 attach to=456"
  }
}
```

---

Alternative to https://github.com/visionmedia/debug/pull/543

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -65,6 +65,24 @@ module.exports = function setup(env) {
 
   function createDebug(namespace) {
     var prevTime;
+    var lastId = 0;
+
+    function prepareArgs() {
+      // turn the `arguments` into a proper Array
+      var args = new Array(arguments.length);
+      for (var i = 0; i < args.length; i++) {
+        args[i] = arguments[i];
+      }
+
+      args[0] = createDebug.coerce(args[0]);
+
+      if ('string' !== typeof args[0]) {
+        // anything else let's inspect with %O
+        args.unshift('%O');
+      }
+
+      return args;
+    }
 
     function debug() {
       // disabled?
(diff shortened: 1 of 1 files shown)


indutny:
I probably prefer this one over #543. Perhaps, it might be worth adding colors to it eventually.

indutny:
The coverage is down, because there are no tests 🤖

Qix-:
I definitely see the benefit of this one over #543, though I think the term `prefix` is more apt. Thoughts?

Qix-:
I see the point now, @indutny - not sure why I wasn't getting it the last time I looked at this.

Question is, why not just use another namespace? `.extend()` was added in #524 for similar reasons - using wildcards in namespaces should do something similar, if not the same.

Also, sorry for taking so long to get back to this.

Qix-:
Going to go ahead and close these two (this and #543) in lieu of https://github.com/visionmedia/debug/pull/544#issuecomment-447018584 in order to clean up a bit - feel absolutely free to comment back here and we can pick up the discussion :)

indutny:
@Qix `extend()` is pretty good! Thank you!

Issue #467, opened 2017-06-09
webpack.DefinePlugin and process.env.DEBUG


wontem:
My webpack config:
```
...
plugins: [
  new webpack.DefinePlugin({
    'process.env': {
      NODE_ENV: JSON.stringify('production'),
      DEBUG: true
    }
  })
],

target: 'electron-main',
...
```

So webpack changes
`process.env.DEBUG = namespaces;`
to
`true = namespaces;`
which leading to error
`ReferenceError: Invalid left-hand side in assignment`

aga5tya:
I too have this issue. @wontem , did you find a solution to this?

eudaimos:
I tried this in Chrome using the following in my client-side `index.js` per the [README suggestion](https://github.com/visionmedia/debug#browser-support) of setting `localStore.debug`
```javascript
if (window.localStorage && process.env.NODE_ENV === 'development') {
  localStorage.debug = process.env.DEBUG;
}
```
and it's still not working

I have this lower in the same file so I know it _can_ work
```javascript
if (process.env.NODE_ENV === 'development') {
  module.hot.accept('./app/App', () => {
    const NextApp = require('./app/App').default; // eslint-disable-line global-require
    render(NextApp);
  });
}
```

eudaimos:
@aga5tya and @wontem and anyone else who comes along here
THIS WORKS!! :)
Using the following in my client-side `index.js` per the [README suggestion](https://github.com/visionmedia/debug#browser-support) of setting `localStore.debug`
```javascript
if (localStorage && process.env.NODE_ENV === 'development') {
  localStorage.debug = process.env.DEBUG;
}
```

FYI - my first attempt checking `window.localStorage` didn't work, so be mindful of that.

@maintainers - this should be closable and maybe add the snippet to the README for the webpackers out there using the DefinePlugin

SPGWhistler:
Although it isn't exactly the same issue, I had an issue where if I didn't specify the DEBUG environment variable, webpack was importing all of my environment variables - with the end result being my build environment variables ending up in my final source code. Not good. The quick and easy solution was to just add DEBUG="" to the environment variables. When webpack builds, it finds it, and just includes the empty string instead of all of the environment variables.

The issue was in browser.js, in the 'load' method, and after webpack did its thing, the output looked like this:
[...]

PR #560, opened 2018-04-18
Fix exposing of .env variables in Create React App 

moos:
Create React Apps have the [nifty feature](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-development-environment-variables-in-env) of getting environment variables either from the shell or `.env` files.  These are then resolved and baked into the JS bundle by webpack.

When using debug, the entirety of your `.env` file gets exposed in the bundle, e.g.:
```js
    function u() {
      var e;
      try {
        e = t.storage.debug
      } catch (e) {
      }
      return !e && "undefined" !== typeof r && "env" in r && (e = Object({
        NODE_ENV: "production",
        PUBLIC_URL: "",
        REACT_APP_FOO: "1"
      }).DEBUG), e
    }
```
This PR fixes that situation by moving the env access to a separate file that isn't accessed when in browser mode.

Also should resolve https://github.com/visionmedia/debug/issues/467#issuecomment-324054986.


Qix-:
I'm :-1: on this. It's adding very specific checks where this is really a shortcoming in electron. Is this even still an issue? I know this is kind of an old PR.

moos:
Yup - definitely still an issue.  It's not an electron issue -- the code referenced above ends up in CRA bundle that's consumed by the _browser_, thereby exposing the **entirety** of CRA app's various `.env` files to the web site visitor -- clearly, an unintended (and _hard-to-detect_, unless you analyze the bundle -- I bet most app devs don't do that!) side-effect of `.env` + `debug`.

I'd be happy to submit an update if there is interest.

Qix-:
Hi, sorry it took so long to get back. Could you rebase please? I'm okay with adding this check.

karma.conf.js
```diff
@@ -14,7 +14,7 @@ module.exports = function (config) {
 		// Test results reporter to use
 		// possible values: 'dots', 'progress'
 		// available reporters: https://npmjs.org/browse/keyword/karma-reporter
-		reporters: ['progress'],
+		reporters: ['mocha'],
```
moos: So we can see that browser env was selected.

package.json
```diff
@@ -25,9 +25,10 @@
   "license": "MIT",
   "scripts": {
     "lint": "xo",
-    "test": "npm run test:node && npm run test:browser",
+    "test": "npm --node run test:node && npm run test:browser && npm --electron run test:electron",
```
moos: `--node` and `--electron` get passed to test as `process.env.npm_config_node` and `process.env.npm_config_electron`.

src/browser.js
```diff
@@ -204,20 +212,12 @@ function save(namespaces) {
...
 	try {
-		r = exports.storage.getItem('debug');
+		return exports.storage.getItem('debug');
 	} catch (error) {
 		// Swallow
 		// XXX (@Qix-) should we be logging these?
 	}
-
-	// If debug isn't set in LS, and we're in Electron, try to load $DEBUG
-	if (!r && typeof process !== 'undefined' && 'env' in process) {
```
moos: This is the critical part that was exposing `process.env` in the browser.  It's been pulled out into the electron-specific flow.

moos:
holy 🐟  -- let's hope I don't have to do that again!

moos:
Ping!  Would hate to see this become stale again.

818Nawaf (approved):

818Nawaf (approved):

PR #569, opened 2018-05-11
Fix nwjs support

miaulightouch:
fix debug won't show message in nwjs

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -7,7 +7,8 @@ exports.formatArgs = formatArgs;
 exports.save = save;
 exports.load = load;
 exports.useColors = useColors;
-exports.storage = 'undefined' != typeof chrome
+exports.storage = 'undefined' == typeof process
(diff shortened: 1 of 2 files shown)


Qix-:
I'm :-1: on this, seems like a very platform-specific check. An issue might need to be opened over at nwjs.

miaulightouch:
I don't get it.
if this is very platform-specific check, then `window.process.type === 'renderer'` should be removed and usage on Electron would be broken.

`process.type` is not presented in normal nodejs, it's only on Electron to show current process is  renderer process or main process. so what's the issue for nwjs?

In the other hand, nwjs redirect all process.stdout, process.stderr to devtools, if let debug work like in nodejs, it would be very tricky.

Qix-:
Hey, you're right. Can you rebase?

Qix-:
Thanks :)

PR #572, opened 2018-05-28
document `DEBUG_MAX_ARRAY_LENGTH`

darky:
This feature exists, but not documented.


TooTallNate:
The `Note:` underneath that table _attempts_ to document it, but apparently doesn't do a very good job.

darky:
Yep, I just noticed, that here link to node.js doc about additional options.
But discouraged situation, that some options described in table.
Better to describe all node.js options or remove `DEBUG_COLORS`, `DEBUG_DEPTH`, `DEBUG_SHOW_HIDDEN`, because it described in node.js doc

PR #579, opened 2018-06-13
Adding TVMLKit support

febbraro:
Apple TV has a JS runtime that allows you to built apps with Javascript.  There are some intricacies of it though. There is no `window` object and you need to access Storage with the `getItem` and `setItem` functions instead of directly accessing properties on the storage object.

This PR adds support for TVMLKit by simply checking if we are in a TVMLKit environment and returning the proper `localStorage` object and by using the item getter/setter calls for the `debug` value.

Just adding some docs related to the storage API and compatibility with `getItem`/`setItem`
https://developer.mozilla.org/en-US/docs/Web/API/Storage/LocalStorage
https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -169,11 +169,8 @@ function load() {
 function localstorage() {
   try {
     // TVMLKit (Apple TV JS Runtime) does not have a window object, just localStorage in the global context
-    if (typeof window === 'undefined' && typeof navigationDocument !== 'undefined') {
-      return localStorage;
-    }
-
-    return window.localStorage;
+    // The Browser also has localStorage in the global context.
+    return localStorage;
   } catch (e) {}
 }
 
--- a/src/browser.js
+++ b/src/browser.js
@@ -129,7 +129,7 @@ function save(namespaces) {
     if (null == namespaces) {
       exports.storage.removeItem('debug');
     } else {
-      exports.storage.debug = namespaces;
+      exports.storage.setItem('debug', namespaces);
     }
   } catch(e) {}
 }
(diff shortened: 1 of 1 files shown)


src/browser.js
```diff
@@ -168,6 +168,11 @@ function load() {
 
 function localstorage() {
   try {
+    // TVMLKit (Apple TV JS Runtime) does not have a window object, just localStorage in the global context
+    if (typeof window === 'undefined' && typeof navigationDocument !== 'undefined') {
+      return localStorage;
+    }
+
     return window.localStorage;
```
TooTallNate: Maybe we could just change this line to `localStorage` and have no `if` check?

febbraro: That would definitely work for the TVMLKit scenario, does that work for the general browser scenario as well?

TooTallNate: `window` is the global object in the web browser, so I can't really think of a scenario where it would be a problem. Only one way to find out though, haha.

Qix-: Yeah I concur. That would work fine.

febbraro: Ok, I'll make that change

febbraro:
Ok, PR updated to remove `window.` from the reference to localStorage.

src/browser.js
```diff
@@ -169,11 +169,8 @@ function load() {
...
   try {
     // TVMLKit (Apple TV JS Runtime) does not have a window object, just localStorage in the global context
-    if (typeof window === 'undefined' && typeof navigationDocument !== 'undefined') {
-      return localStorage;
-    }
-
-    return window.localStorage;
+    // The Browser also has localStorage in the global context.
+    return localStorage;
   } catch (e) {}
```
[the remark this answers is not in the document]
stephenmathieson: Old versions of IE don't support local storage, so this would throw when attempting to access it.

Issue #584, opened 2018-06-22
Migrate makefile to npm


DanielRuf:
I will start the migration away from the makefile to native npm scripts.

PR #585, opened 2018-06-22
chore: cache node_modules

DanielRuf:


Qix-:
Hey there - this broke tests it appears.

DanielRuf:
Well, I guess this is some issue with the makefile. In general the scripts shoul be in package.json and .travis.yml.

I guess I'll port the makefile to npm scripts and .travis.yml scripts.
Because makefiles do not work on all platforms and this is more complex than needed.

Qix-:
GNU Make* doesn't work on all platforms. Makefiles work just fine. MSVC has its own make but it's slightly different.

I agree about moving those into package.json, as per #584.

Qix-:
We also don't include lockfiles in the repository, so I'm not sure how much I like the idea of caching node_modules. IMO it's an anti-pattern caching anything in CI.

DanielRuf:
> IMO it's an anti-pattern caching anything in CI.

In general: no, CircleCI also does it and all others do it too. =)

Qix-:
I understand that the option is there to cache various artifacts, but the whole point of CI is to test - fresh - the entire process, from source checkout to final target artifacts. Caching and re-using those things is an anti-pattern given that definition.

Just because software gives you the ability to do something doesn't necessarily mean it's the best idea to utilize it. Plus, with this module in particular, the performance gains are tiny while the potential for error is high. There are already _so many_ problems with Node.js package managers under _normal_ use that I'm not sure there's a good enough reason to attempt to optimize in a way that might produce false positives/negatives.

@TooTallNate @thebigredgeek thoughts?

DanielRuf:
I see it differently also this is just about the dependencies and I saw no issues so far, also in big projects.

There is a reason for npm ci and the caches, also composer has a cache (for CI).

Qix-:
And what is that reason?

DanielRuf:
> And what is that reason?

For what?

Qix-:
> There is a reason for npm ci and the caches

What is the purpose of a cache in CI other than to decrease _happy path_ builds?

DanielRuf:
Not sure what you mean. You normally do not reinstall the dependencies on every install (on normal machines), there are `npm ci` and lockfiles (they are meant to be committed), ...

Dependencies like jQuery, lodash, upath, ... do not change between builds.

So far I see no real facts against build caches ) and it's also for better DX as we can get more builds in the same time and testing is faster. For example magento/magento2 had definitely faster build. CircleCI 2 per default creates efficient build caches and their builds are extremely fast.

And I still did not see any issues arising in big projects of orgs like facebook/jest, angular/angular and others.

Can you provide real and current reasons not to use any caches on CI (even npm or composer cache) and actual issues of this?

Qix-:
> Not sure what you mean. You normally do not reinstall the dependencies on every install (on normal machines)

- CI is not a normal machine
- CI is not an "install", it's CI. It's testing the entire flow from beginning to end.

> Dependencies like jQuery, lodash, upath, ... do not change between builds.

They do if you don't pin them or have lockfiles.

Lockfiles are an anti-pattern for two reasons:

- They are an artifact of the install/build process; artifacts do not belong in source control (just like binaries do not belong in source control)
- They have a very, **very** lengthy history of bugs and false negatives/positives

> So far I see no real facts against build caches

I just presented several, one of which particularly important in https://github.com/visionmedia/debug/pull/585#issuecomment-400093508: CI and tests are not supposed to be _fast_.

To expand a bit, CI and testing is supposed to be _correct_. Period. Caching modules is not correct. Caching modules does not reflect the flow a normal user would experience on their end when consuming the repository.

There _are_ cases where packages would change from build-to-build. Here's one now:

1. Package A depends on package B; A is configured to cache modules, and A's package.json has all modules pinned (no `^` or `~` prefix)
2. Package B gets an update published; package A maintainers are notified
3. Package A updates B in dependencies block to newest version
4. Package A has a build as a result
5. Package B sees massive flaw in release and reacts by unpublishing since they're [within the 24 hour policy NPM has mandated](https://blog.npmjs.org/post/141905368000/changes-to-npms-unpublish-policy) after the `left-pad` incident.
6. PR comes in for package A - no problem! Modules are cached. CI Passes, PR is merged.
7. Package A bumps as a patch release.
8. Package C, which relies upon package A, has `"package-A": "^1.0.0"` in their dependencies block; users automatically start acquiring the new version of Package A
9. Everything breaks.

This sounds far fetched. It isn't. It's [happened before](https://github.com/chalk/ansi-styles/issues/15#issuecomment-202589714) (we were package B in this scenario).

I'll leave other scenarios as an exercise to the reader. Just remember that CI is supposed to be **complete** and **correct**, _not_ fast. Cached modules violate both ends of that contract.

> and it's also for better DX

Not when you have false positives and negatives causing CI to fail, _especially_ since the details of failures are not reported back to GH. Over the course of maintaining a slew of packages, I've found it rare that someone that submits a PR goes into the failed CI builds to see what the error is and push a fix completely autonomously. In most cases it results in a "I'm not sure why tests are failing" even though the error is _right there_.

Dependency errors are hard to track down, especially when you don't own the caching mechanism. As a maintainer, it's one more thing for me to worry about - **whereas _not_ having a cache means I _never_ have to worry about false negatives/positives**.

> For example magento/magento2 had definitely faster build.

They're a larger repository. 10% faster on `debug` means seconds shaved off. 10% faster with magento (which has a [lot<sup>1</sup>](https://github.com/magento/magento2/blob/2.2-develop/package.json.sample#L11-L41) of [dependencies<sup>2</sup>](https://github.com/magento/magento2/blob/2.2-develop/composer.json#L11-L216) and loads of [tests](https://github.com/magento/magento2/tree/2.2-develop/dev/tests)) means minutes or even, potentially, hours.

The comparison doesn't stick. As well, I'd make the same argument for them - I still firmly believe it's an anti-pattern.

> And I still did not see any issues arising in big projects of orgs like facebook/jest, angular/angular and others.

> And now also yargs/yargs caches them.

They also have hordes of development support at their backs. `debug` has three people, each of which maintain a laundry list of other packages.

> Can you provide real and current reasons not to use any caches on CI (even npm or composer cache) and actual issues of this?

Yes, but I'll leave that as an exercise to the reader - lurk around [npm/npm](https://github.com/npm/npm/issues/) for the slew of issues they had when releasing npm@5 and all of the cache/lock problems people were facing. They lost trust within the node.js community.

Further, whether or not they've caused a slew of errors or not is moot - it breaks the *correctness* of builds. Period.

DanielRuf:
Lockfiles are meant to be committed but well, ok.

Qix-:
Yes, and I explained my point above why that's bad design. It's code smell. Lockfiles are a bandaid patch for a deeper failure of the dependency systems in place.

DanielRuf:
And if a package changes then also the build cache gets invalidated. It will install over loaded cache.

Qix-:
Unless something's changed, that's not true with lockfiles. They bypass hitting the server to shave off HTTP request times. If the package is there, meets the hashing requirements and is within the requested semver range they are considered a cache hit.

Which, again, completely invalidates the "complete" and "correct" points made earlier.

Issue #582, opened 2018-06-20
[RFC] Customizing output format


Qix-:
Ref #573, #565, #553, #456, #453, #402 

Right now it's very clear that `debug` takes a subjective stance on its output format. This is unsuitable for a variety of use-cases, especially the disparity between the `isatty(2) == 0` and `== 1` cases.

Let's figure out how to solve this.

---

My initial knee-jerk idea would be to introduce a `DEBUG_FORMAT` environment variable. It would accept a string with delimiters similar to date-time or printf formats that indicate different elements of a debug log.

We'd have a default format that would then be **the same** between interactive/non-interactive TTY sessions (fd 2, as always).

This would look something like

```
DEBUG_FORMAT="%N %m %D"
```

Where `%N` is the namespace (colored if interactive terminal), `%m` is the message and `%D` is the time diff - the exact format in the [screenshot](https://user-images.githubusercontent.com/71256/29091486-fa38524c-7c37-11e7-895f-e7ec8e1039b6.png) in the [readme](https://github.com/visionmedia/debug#debug).

Of course there would be other delimiters available, and the above doesn't necessarily need to be the default, but it would solve pretty much any use-case imaginable regarding the output format in a non-subjective manner.

---

That's just my personal proposal, but this RFC is after _any_ solution, so if you have a better idea please sound off below.

TooTallNate:
Huge 👍 to this proposal, basically as-is. Env based configuration is the only one that really makes sense to me for this module.

Qix-:
This could also be a reason to remove all but `DEBUG` and `DEBUG_DEPTH` environment variables, too. cc @TooTallNate @thebigredgeek

SCG82:
Here's how I prepend `[ YYYY/MM/DD - HH:mm:ss ]` to all debug output lines:

```javascript
var moment = require('moment');
var debug = require('debug');
debug.formatArgs = formatArgs;
function formatArgs(args) {
    let name = this.namespace;
    let useColors = this.useColors;
    let dateTime = '[' + moment().format('YYYY/MM/DD - HH:mm:ss') + ']';
    if (useColors) {
        let c = this.color;
        let colorCode = '\u001b[3' + (c < 8 ? c : '8;5;' + c);
        let prefix = ' ' + colorCode + ';1m' + name + ' ' + '\u001b[0m';
        args[0] = dateTime + prefix + args[0].split('\n').join('\n' + '                       ' + prefix);
[...]

PR #586, opened 2018-06-25
Add option to disable colors in the browser

leahciMic:
Add an option to allow you to disable colors in the browser.

## Why

We use debug in a few browser projects. Sometimes the output of the console ends up in a text file, and the formatting control characters make it difficult to grok.

## Usage

`localStorage.setItem('debug_colors', false);`

## Notes

I had to add an option to reload the `inspectOpts` in the node environment. This is only used for the tests. Perhaps there is a better way, but I did not find one.

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -42,26 +42,29 @@ try {
  *
  *   $ DEBUG_COLORS=no DEBUG_DEPTH=10 DEBUG_SHOW_HIDDEN=enabled node script.js
  */
+exports.loadInspectOpts = function () {
+  exports.inspectOpts = Object.keys(process.env).filter(function (key) {
+    return /^debug_/i.test(key);
+  }).reduce(function (obj, key) {
+    // camel-case
+    var prop = key
+      .substring(6)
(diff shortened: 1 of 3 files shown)


Qix-:
Your redefinition of `inspectOpts` is a breaking change, and thus won't be accepted. Please try to find another way to do what it is you're trying to do.

src/node.js
```diff
@@ -42,26 +42,29 @@ try {
  *
  *   $ DEBUG_COLORS=no DEBUG_DEPTH=10 DEBUG_SHOW_HIDDEN=enabled node script.js
  */
+exports.loadInspectOpts = function () {
```
Qix-: Your redefinition of inspectOpts is a breaking change, and thus won't be accepted. Please try to find another way to do what it is you're trying to do.

Qix-:
Also, please add docs in the README.

Qix-:
Closing due to inactivity. Also, we might be addressing this in #582.

PR #590, opened 2018-07-13
fix #467

### LINKED ISSUE #467: webpack.DefinePlugin and process.env.DEBUG

errorx666:


errorx666:
As stated in #467, if using webpack's `DefinePlugin` will rewrite `process.env.DEBUG` to a literal `true` or `false`.

The function becomes:
```js
function save(namespaces) {
  if (null == namespaces) {
    // If you set a process.env field to null or undefined, it gets cast to the
    // string 'null' or 'undefined'. Just delete instead.
    delete true; // syntax error
  } else {
    true = namespaces; // syntax error
  }
}
```

With this patch, we avoid assigning to an r-value.

This is currently blocking me from using the bcrypt library.

Qix-:
Please put all of that into a comment, and don't alias as it's confusing (I thought the aliasing fixed something bizarre, hence my previous comment).

errorx666:
I added the comment.

Qix-:
Sorry it took me so long to get to this. Any chance you could rebase off master?

Qix-:
Actually, I've changed my mind on this - please see https://github.com/visionmedia/debug/issues/467#issuecomment-448419976.

Issue #603, opened 2018-09-11
[VISIBILITY] 3.2.0 wrongly released as minor version


mariodu:
remove support for node.js v4 is a breaking change, should publish with major version.

jtyjty99999:
+1

dead-horse:
this module is widely used in node community, it is hard to pin the version of debug in application's package.json.

replace all `let` in https://github.com/visionmedia/debug/commit/ba8a424d41e9dc6129e081ac3aa9715be6a45fbd with `var` can quickly resolve this problem, it doesn't hurts :)

Qix-:
![image](https://user-images.githubusercontent.com/885648/45344916-16dffa80-b562-11e8-839e-7552d67216eb.png)

Side effect of going so long between releases is stuff like this is missed.

Node 4 is out of LTS, and has been for a long while. It was last released in March of this year. The last non-LTS version was released **in 2015**.

Here's what I'll do:

- babel-ify the current version back to the stone ages for those on Node 4
- release as `3.2.2`
- immediately deprecate `3.2.2`, `3.2.1` and `3.2.0` on npm
- release `4.0.0` that will be identical to `3.2.1`

Is that acceptable?

Qix-:
@dead-horse That wouldn't work. There are a whole arsenal of things Node 4 doesn't support.

dead-horse:
https://github.com/visionmedia/debug/issues/603#issuecomment-420176933 seems good. 👍

jtyjty99999:
Complete solution 👍

mariodu:
👍

Qix-:
@mariodu @jtyjty99999 @dead-horse `3.2.2` is released - can you confirm that fixes Node 4 for you?

medikoo:
Other problem is that with `v3.2` `node.js` file was removed. Such change is also a breaking and should not be published with minor.

Still I'm not sure if any modules which explicitly relied on `3.x` are affected.

I was affected but only because I pointed v3 to deps which expected v2 but for which v3.1 was  still safe to use (e.g. `babel-core` which requires `debug/node`)

Qix-:
@medikoo what was `node.js` being used for? It wasn't referenced in any of the meta-information that I could see, hence why it was removed.

medikoo:
> I'll re-introduce node.js in 3.2.3 begrudgingly for those using an outdated babel-core

Thank you!

Qix-:
[...]

PR #607, opened 2018-09-11
Fix browserify and supply alternative unpkg entry point

### LINKED ISSUE #606: debug no longer works with browserify
Any version of debug that I use past version 3.1.0 gives this error when trying to browserify code that includes this module

```
Cannot find module './common' from '[redacted]/node_modules/debug/dist'
      at FSReqWrap.oncomplete (fs.js:154:21)
```

There was a mention in #603 about a browserify incompatibility, but it doesn't appear to be the same issue.

Qix-:
Closes #606.
Ref #603.

This restores Browserify functionality when `debug` is used as a nested module. It also fixes the unpkg use case (though in reality there should be a better, more standard way to do this).

However, it's the least intrusive way to fix both problems. I can't guarantee that the unpkg solution works, but users should be pinning versions anyway and not using `debug@*` on live sites.

cc @mjackson - if you have any input as to how to support unpkg in a less domain-specific way, that'd be great.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ dist: dist/debug.js dist/test.js
 .INTERMEDIATE: dist/debug.es6.js
 dist/debug.es6.js: src/*.js
 	@mkdir -p dist
-	browserify --standalone debug $< > $@
+	browserify --standalone debug . > $@
 
 dist/debug.js: dist/debug.es6.js
 	@mkdir -p dist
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -11,7 +11,7 @@ dist: dist/debug.js dist/test.js
 .INTERMEDIATE: dist/debug.es6.js
 dist/debug.es6.js: src/*.js
 	@mkdir -p dist
-	browserify --standalone debug $< > $@
+	browserify --standalone debug . > $@
```
Qix-: Reverted this to make sure browserify configurations in the future work as intended.

PR #624, opened 2018-10-04
feat: Return namespaces string when invoking disable()

### LINKED ISSUE #523: Feature: disable() returns string of previously-enabled namespaces
So that debugging can be turned off and turned back on without needing to know what was turned on, it would be useful to have disable() return a string containing the list of namespaces that were enabled prior to the disable() call.  The string should be suitable for passing to enable().

mblarsen:
PR for feature request #523 

I thought about adding a new property `createDebug` which would hold the untouched string passed to `enable` but decided against it and instead let the disable deal with what it has: `names`, `skips`.

Let me know if you prefer adding an untouched namespaces string to createDebug instead or you any other changes.

Closes #523

ps: Happy Hacktoberfest

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,24 @@ $ DEBUG=foo node -e 'var dbg = require("debug"); dbg.enable("bar"); console.log(
 => false
 ```
 
+`disable()`
+
+Will disable all namespaces. The functions returns the namespaces currently
+enabled (and skipped). This can be useful if you want to disable debugging
+temporarily without knowing what was enabled to begin with.
+
+For example:
+
+```js
+let debug = require('debug');
+debug.enable('foo:*,-foo:bar');
+let namespaces = debug.disable();
+debug.enable(namespaces);
+```
+
+Note: There is no guarantee that the string will be identical to the initial
+enable string, but semantically they will be identical.
+
 ## Checking whether a debug target is enabled
 
 After you've created a debug instance, you can determine whether or not it is
--- a/test.js
+++ b/test.js
@@ -81,4 +81,43 @@ describe('debug', () => {
 			expect(logBar.namespace).to.be.equal('foobar');
 		});
 	});
+
+	describe('rebuild namespaces string (disable)', () => {
+		it('handle names, skips, and wildcards', () => {
+			debug.enable('test,abc*,-abc');
+			const namespaces = debug.disable();
+			expect(namespaces).to.equal('test,abc*,-abc');
+		});
+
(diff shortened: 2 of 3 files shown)


mblarsen:
Let me add test cases if you want and if you are okay with the implementation so we can get the coverage up again :)

Qix-:
Sure, this is fine. :)

mblarsen:
They were added already https://github.com/visionmedia/debug/pull/624/commits/ac5a726cb09d8b14ebfaa0661fe1a25422626b66#diff-1dd241c4cd3fd1dd89c570cee98b79ddR85

But spotting a spelling error. Will just correct it.

Qix-:
It might seem redundant but can you add a few tests that test passing the returned string back to `.enable()`?

mblarsen:
I added a test of using enable. I'm using the `names` and `skips` to check for equality.

test.js
```diff
@@ -81,4 +81,43 @@ describe('debug', () => {
...
+
+		it('names+skips same with new string', () => {
+			debug.enable('test,abc*,-abc');
+			const oldNames = [...debug.names];
+			const oldSkips = [...debug.skips];
+			const namespaces = debug.disable();
+			expect(namespaces).to.equal('test,abc*,-abc');
+			debug.enable(namespaces);
+			expect(oldNames.map(String)).to.deep.equal(debug.names.map(String));
+			expect(oldSkips.map(String)).to.deep.equal(debug.skips.map(String));
```
Qix-: Why the `.map(String)` here?

mblarsen: They are RegExp objects, so they are not the same even though the regular expression is the same. Mapping to string will make the chai deep equal happy.

Qix-: Sure, sounds good.

Qix-:
Last thing: could you add a bit of documentation to the readme?

https://github.com/visionmedia/debug#set-dynamically

mblarsen:
There you go!

README.md
```diff
@@ -317,6 +317,24 @@ $ DEBUG=foo node -e 'var dbg = require("debug"); dbg.enable("bar"); console.log(
 => false
 ```
 
+`disable()`
+
+Will disable all namespaces. The functions returns the namespaces currently
```
mblarsen: @Qix- I added some documentation here. I don’t know if this is what you wanted?

Qix-: Yes :)

Qix-:
Thank you again :) I'll get this released shortly.

PR #625, opened 2018-10-09
fix: performance issue

gxcsoccer:
image

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -64,13 +64,13 @@ function setup(env) {
     var prevTime;
 
     function debug() {
-      for (var _len = arguments.length, args = new Array(_len), _key = 0; _key < _len; _key++) {
-        args[_key] = arguments[_key];
-      }
-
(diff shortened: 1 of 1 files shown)


src/common.js
```diff
@@ -64,15 +64,15 @@ function setup(env) {
...
-      }
-
       // Disabled?
       if (!debug.enabled) {
         return;
       }
 
+      for (var _len = arguments.length, args = new Array(_len), _key = 0; _key < _len; _key++) {
+        args[_key] = arguments[_key];
+      }
```
gxcsoccer: should not run if debug is disabled

fengmk2:
LGTM

Qix-:
Does this exist in master, too?

gxcsoccer:
> Does this exist in master, too?

@Qix- master is ok

Qix-:
Released as `debug@3.2.6`.

Issue #612, opened 2018-09-14
Include supports-color as dependency


Qix-:
Right now, we enable 256 colors if you have `supports-color` as a dependency. A few individuals have mentioned it should be included in the project dependencies, which sounds ideal.

The reason why we didn't (initially) do so is because we didn't want a minor/patch bump to introduce yet another dependency.

Thoughts?

Jezorko:
I think it's better to give users a choice. Not everyone needs fancy colors. Also, how to add them is well explained in the README.

Qix-:
@jezorko Yeah I'm back and forth, but I mostly agree. A few times, though, people have complained about it - hence this ticket.

PR #628, opened 2018-10-15
Use rollup dist bundle instead. Add esm module distribute for rollup ecosystem.  (#468)

allex:
Use rollup dist bundle instead. Add `dist/debug.esm.js` module distribute for rollup ecosystem.  (#468)

### Use for Rollup ecosystem

```js
import resolveId from '@allex/rollup-plugin-node-resolve'

// rollup.config.js
export default {
  input: 'src/main.js',
  output: {
    file: 'dist/bundle.js',
    format: 'umd'
  },
  plugins: [
    // ...
    resolveId({
      extensions: ['.js', '.ts'],
      alias: {
        'debug': 'debug/dist/debug.esm.js',
      }
    })
  ]
  // ...
}
```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,7 +6,12 @@
         "targets": {
           "chrome": "58",
           "ie": "8"
-        }
+        },
+        "exclude": [
+          "transform-typeof-symbol",
+          "transform-function-name",
+          "transform-classes"
+        ]
       }
     ]
   ]
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,38 @@
+// Type definitions for debug
+// Project: https://github.com/visionmedia/debug
+// Definitions by: Seon-Wook Park <https://github.com/swook>
+//                 Gal Talmor <https://github.com/galtalmor>
+//                 John McLaughlin <https://github.com/zamb3zi>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare var debug: debug.IDebug;
+
+export = debug;
+export as namespace debug;
+
+declare namespace debug {
+    export interface IDebug {
+        (namespace: string): debug.IDebugger,
+        coerce: (val: any) => any,
+        disable: () => void,
+        enable: (namespaces: string) => void,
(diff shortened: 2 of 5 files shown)


Qix-:
Hi there - before I really comment on this at a high level, please make sure all of the formatting is consistent. This also looks like it adds a lot of superfluous stuff that doesn't need to be there. Could you make sure you're adding the minimal amount necessary to build with rollup?

Also, I'm still not sold on the idea of rolling up modules. That's been an area of extensive discussion over at `chalk`, too.

allex:
> Hi there - before I really comment on this at a high level, please make sure all of the formatting is consistent. This also looks like it adds a lot of superfluous stuff that doesn't need to be there. Could you make sure you're adding the minimal amount necessary to build with rollup?
> 
> Also, I'm still not sold on the idea of rolling up modules. That's been an area of extensive discussion over at `chalk`, too.

Actually, for rollup ecosystem, we need bundle a out dist with ES module syntax. so i've dropd bundler of `browserify` and use `rollup` instead.

Qix-:
This is being closed - reasoning to follow. However, _for the future_:

- Why are you introducing typescript definitions in this PR? They are entirely unrelated.
- Please don't add your name to the contributors key. We don't use that.
- Please address https://github.com/visionmedia/debug/pull/628#issuecomment-447016548
- Please don't add "added by ..." comments to everything. That's what Git is for.
- Please don't add Vim formatting comments

---

At a high level, I don't see a benefit of this. Really. This is just adding to the build process and doubling the size of a package that garners **millions** of downloads.

My stance on the pre-babelification of modules is that of Sindre Sorhus's: https://github.com/sindresorhus/ama/issues/446

We already deprecated the build process in 4.x. Babel is ultimately going away. This is moving us backward at this point.

Furthermore, `debug` is not that complicated. It has pretty much no bootup cost and [is already pretty cheap](https://bundlephobia.com/result?p=debug@4.1.0). Further when #612 is added, it will already add some size - so I want to minimize the extra cruft as much as possible.

PR #634, opened 2018-11-05
Update readme, for clarity on "*" convention in name

GioLogist:
The following convention only works on node currently. Noticed that while testing in browser and figured i'd update the doc.

> If you append a "*" to the end of your name, it will always be enabled regardless of the setting of the DEBUG environment variable


GioLogist:
@Qix- Sure thing. I wouldn't go as far as removing completely, since Node folks can still enjoy. But def would be nice to have on web 😅

PR #645, opened 2018-12-13
Custom output formatting

### LINKED ISSUE #582: [RFC] Customizing output format

chagris:
Closes #582.

This pull request is an attempt at implementing https://github.com/visionmedia/debug/issues/582. As stated by [my comment](https://github.com/visionmedia/debug/issues/582#issuecomment-446783778) there, this code isn't yet complete and actually ready to be merged, and is waiting for edits and feebacks!

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -43,10 +43,10 @@ function setup(env) {
 	createDebug.outputFormatters = {};
 
 	/**
-	 * Map %m to outputting diff
+	 * Map %m to applying formatters to arguments
 	 */
 
-	createDebug.outputFormatters.m = function(format, args) {
+	createDebug.outputFormatters.m = function (_, args) {
 		args[0] = createDebug.coerce(args[0]);
 
 		if (typeof args[0] !== 'string') {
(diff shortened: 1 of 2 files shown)


chagris:
I added the "metaformatters" flag prefixes `%J` (JSON.stringify()), `%c` and `%C` (color and color + bold (node)) as I proposed here https://github.com/visionmedia/debug/issues/582#issuecomment-446973503, however I'm not 100% on it (they could just as well stay special-case-flags), so it should be easy to revert

Also, I'm not sure what to do about [node's formatters](https://github.com/visionmedia/debug/blob/master/src/node.js#L246) `%o` / `%O` coloring (based on `DEBUG_COLORS` and `isatty()`), and passing other `inspectOpts`

Qix-:
Hey @chagris - sorry I haven't had time to review this or give much input. However I appreciate the work being done.

Admittedly it'll be a little longer before I'll be able to get to this; however, I've done two things here:

1. I revised this PR's title a bit to be less 'prototype'-ey
2. I added the magic keywords `Closes #xxx.` which Github will handle upon merging this PR.

RabiaSaid:
2.6.8

RabiaSaid:
4.3.4

818Nawaf (approved):

818Nawaf (approved):

Issue #665, opened 2018-12-24
Bug: When loaded inside node with --experimental-modules no colors


frank-dspeed:
node --experimental-modules index
```js
// index.mjs
import Debug from 'debug';
import './other.mjs'
Debug('viewModel=>')('init')
```

```js
// other.mjs
import Debug from 'debug';
Debug('other=>')('init')
```

both return with the same color in my case always red

Qix-:
That's a coincidence. The namespaces are hashed based on the name. Install `supports-color` as a dependency to get more color.

PR #666, opened 2018-12-25
Update README.md

frank-dspeed:
Made Clear what a namespace is in the projet Terminology as many people like me would think that the prefix is the namespace

```js
const myDebuger = require('debug')('namespace')
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,11 @@ npm script example:
 
 Every debug instance has a color generated for it based on its namespace name.
 This helps when visually parsing the debug output to identify which debug instance
-a debug line belongs to.
+a debug line belongs to. The Namespace is the Object Name that u assign the debug instance to
+
+```js
+const myDebuger = require('debug') //=> myDebugger is the namespace
+```
 
 #### Node.js
 


Qix-:
This is entirely incorrect. I'm not surr what's confusing you but the namespace is what you pass to Debug initially. The example is very clear.

Also, please don't use "u" as a replacement for "you" in technical literature. Ever.

frank-dspeed:
@Qix- your confusing me as i told you https://github.com/visionmedia/debug/issues/665 in this issue that it did Not Change the output color based on what i supplyed inital to it

Qix-:
Please re-read what I wrote in that comment. I was very clear.

frank-dspeed:
@Qix- ok i should stop to inform you about stuff i understand but i was also very clear your simply wrong as my terminal supports more colors then red and the problem is Not a coincidence.
but anyway i like this lib its ok for me i can use it without colors i only wanted to report that 

i hope that now made sense for you

Qix-:
@frank-dspeed Please **re-read** what I wrote. It is a coincidence that both of those namespaces are red. Try other namespaces or install `supports-color` to support more colors. I don't know how to spell it out any other way.

PR #671, opened 2019-01-07
fix IE compatibility

### LINKED ISSUE #668: it can not work on IE?
image
image
image

swimyoung:
fix #668 

It is because that IE doesn't support arrow function. I've changed arrow function to function expression.


Qix-:
Nope, please see reasoning here:

https://github.com/sindresorhus/ama/issues/446#issuecomment-281014491

Please transpile your code if you need it for old browsers (IE for example).

PR #672, opened 2019-01-10
Remove es6 code from the browser version

### LINKED ISSUE #668: it can not work on IE?

artemave:
To make it IE compatible.  Fixes #668 

EVERYBODY HATES IE. Especially those stuck supporting it. Let's be kind. Our scorn is the last thing they need.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,10 +25,11 @@
   "license": "MIT",
   "scripts": {
     "lint": "xo",
-    "test": "npm run test:node && npm run test:browser",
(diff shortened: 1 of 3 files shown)


Qix-:
Hi - this package uses ES2015+ on purpose. It is intended that you compile your code to work in the range of browsers that is right for your application using a tool such as [Babel](https://babeljs.io/).

For more information: https://github.com/sindresorhus/ama/issues/446#issuecomment-281014491

Thank you for your contribution.

PR #677, opened 2019-01-14
- use simple for loop instead of forEach

knalbandianbrightgrove:
Babel is not transforming on runtime such things like forEach which is critical when a product going to be used as SaaS solution.
I have everything to transpile ES6 code, but some specific methods make me sad.
I would appreciate if you approve this small change

Thank you in advance!

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -14,7 +14,7 @@ function setup(env) {
 	createDebug.humanize = require('ms');
 
 	for (let i = 0; i < Object.keys(env).length; i++) {
-		let key = Object.keys(env)[i];
+		const key = Object.keys(env)[i];
 		createDebug[key] = env[key];
 	}
 
--- a/src/common.js
+++ b/src/common.js
@@ -13,9 +13,10 @@ function setup(env) {
 	createDebug.enabled = enabled;
 	createDebug.humanize = require('ms');
 
-	Object.keys(env).forEach(key => {
+	for (let i = 0; i < Object.keys(env).length; i++) {
+		let key = Object.keys(env)[i];
 		createDebug[key] = env[key];
-	});
+	}
 
 	/**
 	* Active `debug` instances.


Qix-:
`.forEach` has been around since ES5. Your loop is also incredibly inefficient, pulling `Object.keys()` each iteration.

There is no reason why this should be merged, nor should the existing code be causing you problems. If Babel is transpiling this incorrectly, please open an issue with them.

knalbandianbrightgrove:
@Qix- 
Thank you for the quick response.

I made a type when talk about ES6, originally I mean ES5.

Babel is not transpiling .forEach, it should be polyfilled. So in order to prevent extra polyfilled 

One more argument is that in this file you already using the same idea, please take a look
https://github.com/knalbandianbrightgrove/debug/blob/88584c3ecbcee0447e4ff37018085106ec34ef2b/src/common.js#L184

Qix-:
It's not transpiling down to es3? What ancient browser are you targeting?

knalbandianbrightgrove:
@Qix- it does transpiling everything but not methods, in order to work with such things like Array.prorotype.forEach you should use babel-polyfill. I need to support IE8+.
I do not see any performance downgrade in switching to for loop if you care about performance

knalbandianbrightgrove:
@Qix- sorry for disturbing you.

Any logic reason to not approve this PR?

Qix-:
Babel 7 includes support for this, your PR is for compatibility reasons that are caused by user error. It doesn't make any improvement to me. Please open an issue with Babel if polyfilling isn't working as expected.

knalbandianbrightgrove:
@Qix- Please refer to this line https://github.com/knalbandianbrightgrove/debug/blob/88584c3ecbcee0447e4ff37018085106ec34ef2b/src/common.js#L184

And change it to forEach

PR #680, opened 2019-01-28
Add engines to package.json

vkarpov15:
I got bitten today because there isn't really a way to tell that debug 4.x doesn't support Node.js 4. Thanks for all your great work on this module!


christianbundy:
Looks great to me, I'd love to see this merged.

Issue #678, opened 2019-01-20
memory leak when instance is created inside a function.


overflowz:
Hi, I just noticed that when you create debug instance in a function, it is starting to leak the memory without freeing. Here's how you can reproduce it:

```js
const debug = require('debug');

const loop = () => {
  const d = debug('namespace:that:i:want:for:this:function');
  d('hello world');
  setImmediate(loop);
};

loop();
```

If you run this and look at memory, it is leaking a lot without freeing them.
also does not matter if I set environment to DEBUG=* or not, it still leaks. Any thoughts?

EDIT: tested on 3.1.1 and 4.1.1 as well (had 3.1.1 version and then I upgraded to latest one to check if it was fixed).

EDIT2: using node version 10.13 and Windows 10 x64.

Qix-:
This isn't a typical use-case for `debug`, though.

However, I agree that this shouldn't be leaking. I'll flag it as a bug.

overflowz:
well, I found myself pretty much comfy to create instances within a functions, it helps to understand logs much better though (for me), like what function is doing what, etc. But yeah, leaks should not be happening at any cost, dunno.

Qix-:
No, that wouldn't solve it. This will get solved with #645 as the only reason why we're storing references to debug instances is to enable/disable them after they've been created.

I think there's a solution to this in the works, just not directly.

I'll leave this open for now. Good catch, though.

overflowz:
Yeah, I saw that and there were no other usages for it but enabling/disabling the instances.

about #645, I'm not sure, though. I'd either remove the saving part completely (which will remove enable/disable feature) or save instances uniquely by namespace (IMO it is more reasonable to have unique namespaces, since having same namespace in different places could be misleading, dunno).

Qix-:
> save instances uniquely by namespace

overflowz:
Oh, didn't knew about the dynamic part (not using it, hehe). So yeah, will be waiting for them.

Cheers.

brandonros:
> Just found out, that you have to call destroy() on it if you want it to not leak.
> 
> I'm not sure how good this case is but, at least it should be documented like big red fancy text or something since I lost quite a lot of time to understand why my app was leaking though.

Is there anyway we can get https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/debug/index.d.ts updated to expose the `destroy` method?

Qix-:
@brandonros feel free to open a PR there.

[...]

PR #699, opened 2019-04-22
Improvement of memory management for #678

victor-cr:
Avoiding recreating logger for the namespace which has been already created. Fix memory leak for a fixed number of categories. For dynamic categories method destroy is still required. #678

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -19,8 +19,9 @@ function setup(env) {
 
 	/**
 	* Active `debug` instances.
+	* @type {Object<String, Function>}
 	*/
-	createDebug.instances = [];
+	createDebug.instances = {};
 
 	/**
 	* The currently active debug mode names, and names to skip.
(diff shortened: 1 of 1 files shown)


randycoulman:
Thanks for this PR!  I was thinking of submitting a similar one myself, as we just discovered that this memory leak was hitting us hard.  

I was thinking about using a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead of an object for the instances.  Would that be a better choice?

Map seems like the more suitable data structure for this usage and would simplify the code in `destroy` and `enable`, but I'm not sure of the relative performance characteristics, or whether Map is available in every target environment.

victor-cr:
I have no objection of using `Map` - the reason I was using object just to use as simple entity as it is possible. Also, `destroy` and `enable` are rarely used in comparison to logging.

818Nawaf (approved):

PR #707, opened 2019-05-22
Fix enabled method when name ends with *

vhf:
I noticed an issue and I'm pretty sure it's a bug. Here's a PR solving my issue and adding tests for this bug.

The issue: `debug.enabled('foo:*')` is always true, because `name` ends with `*`.

If anyone runs into this issue before it is fixed, there is a workaround: `debug.enabled('foo:*,')`, this does the trick.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -209,7 +209,7 @@ function setup(env) {
 	* @api public
 	*/
 	function enabled(name) {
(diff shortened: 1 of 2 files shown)


vhf:
Ping @Qix- :)

PR #709, opened 2019-06-01
Add `supports-color` as optional peerDependency

jgierer12:
This is the proper way to declare this type of dependency.

Note that if `peerDependenciesMeta` is not supported by the package
manager (yarn < 1.13 and npm) and `supports-color` is not installed, a warning will be
displayed.

See [`peerDependencies`](https://yarnpkg.com/en/docs/dependency-types#toc-peerdependencies), [`peerDependenciesMeta`](https://github.com/yarnpkg/yarn/blob/master/CHANGELOG.md#1130).

Without this, Yarn PnP to throws an error when `debug` is in the dependency tree, since PnP is very strict about `require`ing packages that aren't declared as dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,5 +47,13 @@
     "xo": "^0.23.0"
   },
   "main": "./src/index.js",
-  "browser": "./src/browser.js"
+  "browser": "./src/browser.js",
+  "peerDependencies": {
+    "supports-color": "^6.1.0"
+  },
+  "peerDependenciesMeta": {
+    "supports-color": {
+      "optional": true
+    }
+  }
 }
--- a/src/node.js
+++ b/src/node.js
@@ -23,8 +23,6 @@ exports.useColors = useColors;
 exports.colors = [6, 2, 3, 4, 5, 1];
 
 try {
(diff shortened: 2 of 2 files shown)


eps1lon:
Quick update that `peerDependenciesMeta` is implemented in `yarn@1.13.0`, `npm@6.11.0` but not [yet](https://github.com/nodejs/node/pull/29273) bundled with node and `pnpm@?`

I'm currently not using this feature myself until node bundles `npm@6.11`. If there are any concerns about this feature please let me know.

[original RFC](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md)

arcanis:
Note that the situation changed since the last comment. The `peerDependenciesMeta` is supported by all modern package managers (at least 6 month old). 

Additionally, listing `peerDependenciesMeta` is enough for the peer dependency to be inferred even without the `peerDependencies` field being set - providing a smooth adoption path even for older package managers.

Qix-:
Published as `4.2.0` under the `beta` dist-tag. Please test and make sure nothing has broken for you before I send to `latest`.

PR #715, opened 2019-07-05
Use util.inspect to format stack traces on Node.js

### LINKED ISSUE #711: Extending error object in express, but Debug does only show the stack trace
Hello,

Small question of which I don't know it's a bug or expected behaviour:

In my express application I have the following app.use at the bottom:

```
app.use( (error, req, res, next) => {
    const status = error.statusCode || 500;
    const source = error.source || 'Unknown';
    const message = error.message;
    const info = error.info || '';
    console.log(error); // For logging on the server
    res.status(status).json({
        status: status,
        source: source,
        message: message,
        info: info
    })
});
```

This shows me the stack trace and the error object (with source and info properties).

When I change the **console.log** to **debug**, it only shows the stack trace and not the error object. 

Why is that? And how can I show the error object with the debug command also?

RAnders00:
This patch ensures we don't manually extract the `stack` or `message` from errors if we are running on Node.js. Instead, `util.inspect`/`util.format` does this for us automatically. The browser behaviour remains unaffected.

I added a unit test to verify the behaviour.

Fixes #711.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "karma-mocha": "^1.3.0",
     "mocha": "^5.2.0",
     "mocha-lcov-reporter": "^1.2.0",
+    "sinon": "^7.3.2",
     "xo": "^0.23.0"
   },
   "main": "./src/index.js",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "karma-mocha": "^1.3.0",
     "mocha": "^5.2.0",
     "mocha-lcov-reporter": "^1.2.0",
-    "sinon": "^7.3.2",
     "xo": "^0.23.0"
   },
   "main": "./src/index.js",
--- a/src/node.js
+++ b/src/node.js
@@ -235,7 +235,7 @@ function init(debug) {
 }
 
 /**
- * Coerce `val`. In Node.JS, no values need to be coerced (errors are already being formatted for us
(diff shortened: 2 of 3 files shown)


Qix-:
-1 on the extra dependency just for one test, especially since I disagree with sinon's style of testing.

Without fake timers, you can use regex's to perform the match. You can also patch `log.log` directly (that's what it was built for).

Also, assigning to `.stack` isn't (or hasn't historically been) reliable, as `.stack` is oftentimes dynamically built up via a getter to save expensive string operations at runtime that may not be used.

Further, it's not even specified - you should be using `.toString()` as per section 19.5.1 of the ecmascript standard.

RAnders00:
I was going to test using `startsWith() || endsWith()` initially, but figured that might be too fuzzy. Would that be an acceptable compromise for you?

Qix-:
Why `startsWith` and `endsWith`? Why not a regex?

You're right, nothing we can do about that since Node is apparently non-standard.

RAnders00:
@Qix- Any feedback on the PR? I think the current state resolves your concerns, unless you have further input of course.

calvintwr:
@RAnders00 I want to thank you for this. I have incorporated this into https://github.com/calvintwr/debug/commit/adec948df6e956c2d27ad38d89e69e134dc260a9

Issue #719, opened 2019-07-25
Support decoration with process information


dbo:
When I am running multiple background processes emitting debug logs, it would be helpful if each process could be instructed to emit a process title or its pid when logging.

Qix-:
This falls very cleanly in #582 :) Closing in favor of that.

Issue #556, opened 2018-03-08
Feature check: global log


jasonpincin:
Would you consider a PR that implemented support for globally defaulting `log` to something other than `process.stderr`? The way I’d expect this to work is debug would default to stderr unless `process[Symbol.for(‘DEBUG_LOG’)]` was set. Of course the local `self.log` would always take precedence. 

The motivation for this feature is to allow an application to redirect all debug output to a logger of its choice, without knowing which all dependencies utilize debug. 

I’ve chosen a global Symbol to circumvent the issue of multiple versions of debug in codebase. 

Happy to submit this. Thoughts?

Qix-:
It might be much more practical to just have a top level function on the `debug` instance. However, it's usually an anti-pattern to modify a package's state in a way that affects other packages.

`EventEmitter`s are a good candidate here, too.

```javascript
const debug = require('debug');

debug.on('log', msg => writeToSomeStream(msg));
```

jasonpincin:
The problem with both of these solutions, is I need to know/iterate all of my dependencies that support `debug` and set up a listener or call a function on each of their debug instances. 

The goal is to re-direct _all_ `debug` output to a stream other than `stdout`. An example of why this is useful: to wrap log statements in a structured document so they're coalesced in a log management system, rather than having the messages parsed by line, which makes the `debug` output completely unreadable in any system other than console or a console redirect.

Qix-:
Right now it's going to `stderr` FWIW. Also, why not just use shell redirection?

I still think a top-level function that receives all of the raw output (that defaults to `console.error()`) would solve your problem.

jasonpincin:
Yes, `stderr`, thank you. Shell redirection offers no way to correlate multiple lines. For example, if one `debug.log` call results in 3 lines being written to `stderr`, I have no way of correlating them there. If I can redirect in-process, to something like `Bunyan`, then all three lines are correlated by virtue of the fact that they're in a single structured (JSON) document by the time they reach `stderr` or `stdout`. 

If I have 3 modules, all using `debug`, that I authored, and I `require` them into my application - how do I instruct all three modules to output to my top level function?

I can appreciate that you think it's an anti-pattern; however, all I'm trying to do is control how my application handles debug output. If anything here is an anti-pattern, it's the fact that a dependency is outputting to `stderr` and there's nothing my application, which should govern all I/O, can do about it. 

[...]

PR #720, opened 2019-07-25
#719 - Support decoration with process information

dbo:
Implements #719 

Feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,8 @@ change the behavior of the debug logging:
 | `DEBUG_COLORS`| Whether or not to use colors in the debug output. |
 | `DEBUG_DEPTH` | Object inspection depth.                    |
 | `DEBUG_SHOW_HIDDEN` | Shows hidden properties on inspected objects. |
+| `DEBUG_PROCESS` | Prefixes logs with a title in square brackets. |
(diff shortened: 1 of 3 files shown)


Qix-:
Really not a fan of adding more environment variables - this is what #556 is aimed to solve. Otherwise, every logging combination under the sun would scope-creep this package into oblivion.

dbo:
@Qix- Apparently it's not really about my own logs only (which I could customize with what is aimed in #556), but about 3rd party NPMs that I use in my process. Moreover, environment variables don't require to execute any custom code which is sometimes not possible when plugging into frameworks that use their own startup scripts. Since all debug env variables are scoped to `DEBUG_` I don't see any collision.

Qix-:
It's not about collision, it's about API cleanliness. The idea behind #556 is that it'd be globally enforced and entirely backwards compatible. You wouldn't have to modify code, but instead use `-r` to inject a non-default reporter.

In the case of PID's, #582 could be used here as well.

dbo:
@Qix- I agree, reading on #582 I think it's what I look for, since I could streamline all logs to carry the PID (needs some `%P` or whatever extension though) or define a readable process title.

PR #728, opened 2019-11-01
Adding support for ReactNative

jrobber:
ReactNative renders in a document-less mode, but goes through the browser for their live builds.
Fortunately they tag the navigator as ReactNative making this easy to check for.


jrobber:
With this change you can set the environment variables before you import debug anywhere (like in index.js or in a config.js) like so :
```
process.env.DEBUG = ['*'].join(','); //init,search,addListing:price,reservation:*
process.env.DEBUG_COLORS = true;
```

You could tie those to variables in the manifest if you want to, but the main value of this in ReactNative is development debugging, not distributed.

This will only allow the logs to be colored in the browser terminal.  React-native/expo need to make changes that convert browser colors to server colors if debug is going to work in their terminals.

818Nawaf (approved):

PR #732, opened 2019-12-15
Update common.js

J-Tech-Foundation:


J-Tech-Foundation:
just so if it returns false, there is still the ability to use it without
`
debug.enabled = true
`

Qix-:
That doesn't do what I think you think it does.

PR #738, opened 2020-01-12
README: document debug instance destroy() method

ibc:
Related to #678, this PR adds a new section "Destroying a debug instance" in the README explaining that dynamically created debug instances must be destroyed in order to avoid a memory leak.

It also does a minor/cosmetic cleanup in the README file.


Qix-:
No thanks. #678 is a bug. The readme shouldn't document how to workaround bugs.

PR #740, opened 2020-01-12
Fix memory leak when creating debug instances

### LINKED ISSUE #678: memory leak when instance is created inside a function.

Qix-:
Closes #678 .

One possible solution to the memory leak and needless `.destroy()` method.

Will leave open for a while to allow people to test.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -9,6 +9,16 @@ exports.save = save;
 exports.load = load;
 exports.useColors = useColors;
 exports.storage = localstorage();
+exports.destroy = (() => {
+	let warned = false;
+
+	return () => {
+		if (!warned) {
+			warned = true;
+			console.warn('Instance method `debug.destroy()` is deprecated and no longer does anything. It will be removed in the next major version of `debug`.');
+		}
+	};
+})();
 
 /**
  * Colors.
--- a/src/common.js
+++ b/src/common.js
@@ -12,6 +12,7 @@ function setup(env) {
 	createDebug.enable = enable;
 	createDebug.enabled = enabled;
 	createDebug.humanize = require('ms');
+	createDebug.destroy = destroy;
 
 	Object.keys(env).forEach(key => {
 		createDebug[key] = env[key];
@@ -114,6 +115,7 @@ function setup(env) {
 		debug.useColors = createDebug.useColors();
 		debug.color = createDebug.selectColor(namespace);
 		debug.extend = extend;
+		debug.destroy = createDebug.destroy; // XXX Temporary. Will be removed in the next major release.
 
 		Object.defineProperty(debug, 'enabled', {
 			enumerable: true,
@@ -241,6 +243,14 @@ function setup(env) {
 			return val.stack || val.message;
 		}
 		return val;
+	}
+
+	/**
+	* XXX DO NOT USE. This is a temporary stub function.
+	* XXX It WILL be removed in the next major release.
+	*/
+	function destroy() {
+		console.warn('Instance method `debug.destroy()` is deprecated and no longer does anything. It will be removed in the next major version of `debug`.');
 	}
 
 	createDebug.enable(createDebug.load());
--- a/src/common.js
+++ b/src/common.js
@@ -18,11 +18,6 @@ function setup(env) {
 	});
 
 	/**
-	* Active `debug` instances.
-	*/
-	createDebug.instances = [];
-
-	/**
 	* The currently active debug mode names, and names to skip.
 	*/
 
@@ -63,6 +58,7 @@ function setup(env) {
 	*/
 	function createDebug(namespace) {
 		let prevTime;
+		let enableOverride = null;
 
 		function debug(...args) {
 			// Disabled?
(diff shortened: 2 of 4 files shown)


ibc:
It works fine :)

Will you release this in v4?

Qix-:
Yes, this would be a minor release if pushed.

ibc:
"If pushed"? :)

What does it depend on? I've already changed my projects 3 times today to conform to this issue :)

Qix-:
@ibc 

image

Making sure I didn't make a mistake and break millions of users. Don't expect this to be released in the coming week.

ibc:
Sure, thanks.

iFwu:
I have encountered this problem in the production environment. So I sincerely hope this can be merged ASAP.

ibc:
Which is the state of this PR? is it expected to be merged soon? I already tested it and works fine, I understand it's not trivial to move this into production release but, what is the current state or plan?

Qix-:
> Which is the state of this PR?

image

ibc:
Thanks, somehow I already see it's still open :)
I just mean, is there any ETA or plan to merge it?

ericf89:
Just another follow up that the no-leak branch solved some memory issues I was seeing with a project.

provinceinnovation:
Another follow up - This branch has just fixed the memory leak that's been haunting me for the past 6 months

Qix-:
Published as `4.3.0` under tag `beta` (therefore it's not going to be automatically pulled unless you explicitly ask for it). Please let me know that it still works for you.

ab-pm:
Fixing memory leaks is of course important, but I wonder whether this merge is a performance regression. Has it been tested?
Disabled debug instances previously just checked a single boolean property and then returned early, executing an object lookup and a single branching. Now, they are calling a getter, which in turn calls `createDebug.enabled`, which loops through all the enabled and disabled namespaces and runs a regex `test` on them. This seems much costlier, or am I missing something?

Qix-:
@ab-pm You're exactly right. v5 is going to be a rewrite that will improve this a bit.

`debug` has always been a performance suck, though. Even if the debug namespace is disabled, the parameters are still evaluated going into it. In trivial cases this is fine, but I know for a fact many people are passing expressions to `debug()`. Further, Javascript engines do not have dead code removal advanced enough to be able to determine that pure expressions passed to a no-op function can be elided to safe on cycles.

As for whether or not it's a "performance regression", by this logic anything that adds an extra instruction to the overall code path is a "performance regression". The trade-off here is that `debug` was not correct and leaked memory.

The real fix here would be to overhaul the enabled namespaces system altogether, which would be a major breaking change and is thorougly discussed elsehwere on the issue tracker. As I mentioned before, it's slated for v5, whenever I get around to that.

ab-pm:
@Qix Well it's not just "an extra instruction", but quite many of them, on what might be a hot code path :-) But if you've consciously made the trade-off to favour memory safety, I'm totally fine with this - that's all I wanted to know, thanks!

Qix-:
@ab-pm Right, I understand. However, a branch is always going to be necessary, and if the branch is truthy then I/O is going to be performed - further, _really slow_ I/O (console output is horrendously slow).

My point was more, if you're really worried about the overall performance impact, sprinkling your code with permanent debug logs is not going to help you regardless of the library. Putting any sort of branch in a "hot" code path is going to cause a performance downgrade.

Another thing to consider is that it's not guaranteed a hot codepath is JIT'd, so therefore you're not going to get as much of a benefit from the CPU's pipelining and branch prediction than you would, say, a C/C++/Rust program. If I could, I would mark the branches for `if (namespace_is_enabled)` as "unlikely" (e.g. GCC's `__builtin_expect(namespace_is_enabled, 0)`) to inform the CPU that the branch probably won't hit, and to eat the extra cost of a mispredict in cases where namespaces _are_ enabled (since `debug` **is not** a logging library, despite troves of people using it as such).

So even if it were JIT'd, I'm not sure you'd get a lot of benefit anyway - and further, since it's probably _not_ JIT'd, the branches are going to be pretty equally expensive - and since there is an unavoidable branch here anyway, performance isn't really at the forefront of consideration to the extent that memory incorrectness/leaking is a justifiable tradeoff.

That was my line of thinking :)

cmotsn:
Would a backport of this memory fix leak to 3.x.x version be possible? Some packages do not want to upgrade to debug:4.x in order to not lose compatibility with earlier-but-apparently-still-used versions of Node.js...

Issue #747, opened 2020-02-23
Unable to take control over selectColor


profiprog:
There is possible to reassign `createDebug.selectColor()` method but is not used. Instead that is used directly default definition based on hash calculation.
I would like to have possibility to customise color selection per namespace.

Qix-:
Published as `4.2.0` under the `beta` dist-tag. Please test and make sure nothing has broken for you before I send to `latest`.

PR #748, opened 2020-02-23
Fixes: Unable to take control over selectColor #747

profiprog:
Simple fix


Qix-:
Good catch :) Thank you!

Qix-:
Published as `4.2.0` under the `beta` dist-tag. Please test and make sure nothing has broken for you before I send to `latest`.

PR #749, opened 2020-02-25
Exclude unnecessary files from the npm package

pkuczynski:
Listed a few files and folders which are currently included in the published npm package, while they should not be.


Qix-:
This isn't needed since we use the `files` key in `package.json`. Thanks though!

pkuczynski:
I was looking at the older version 3.1.0, sorry :)

PR #755, opened 2020-04-06
fix quoted percent sign

bmacnaughton:
Inserting `%%` in a message string results in `%%` in the formatted string. This change allows `%%` to work as documented resulting in a single percent sign in the output string.


Qix-:
Good catch, thank you :)

linchao001:
good

Qix-:
Thanks again :)

Issue #766, opened 2020-06-23
% symbol in object interpreted as formatter


matthieusieben:
```js
const debug = require('debug')('foo')
debug.enabled = true // Or use ENV variables
debug("%o", { "%j": "%j %j %%" }, 1, 2, 3)
```

result:
```
  foo { '1': '2 3 %' } +0ms
```

expected result:
```
  foo { "%j": "%j %j %%" } 1 2 3 +0ms
```

This gets even weirder when running the following several times:
```js
debug("%o", { "%j": "%j %j %%" })
debug("%o", { "%j": "%j %j %%" })
debug("%o", { "%j": "%j %j %%" })
debug("%o", { "%j": "%j %j %%" })
```
result:
```
  foo { '"\u001b[38;5;148m+3s\u001b[0m"': '%j %j %' }
  foo { '"\u001b[38;5;148m+847ms\u001b[0m"': '%j %j %' }
  foo { '"\u001b[38;5;148m+674ms\u001b[0m"': '%j %j %' }
  foo { '"\u001b[38;5;148m+620ms\u001b[0m"': '%j %j %' }
```

Am I doing something wrong or is there a way **not** to interpret those `%` symbols?

Qix-:
Reproduced; this is very clearly a bug.

Sorry you're hitting it - I'm surprised it's not been brought up before. Thank you for the reproduction steps.

PR is welcome, else I will try to find time within the next week to get to it.

Qix-:
Super weird. Just investigated a bit more to make sure this wasn't introduced recently (it wasn't, it exists pre-4.x).

It appears, for whatever reason, the `%o` is honoring the second argument as _another_ formatting string.

If you would like to work around this, insert another `%o` at the beginning of the argument list. I know this isn't ideal, but if you need a fix _right now_, that should get you somewhat sane output.

```javascript
const debug = require('debug')('foo');
debug.enabled = true // Or use ENV variables

const someObjectIWantToLog = { "%j": "%j %j %%" }
debug("%o", "%o", someObjectIWantToLog, 1, 2, 3)
debug("%o", "%o", { "%j": "%j %j %%" })
debug("%o", "%o", { "%j": "%j %j %%" })
debug("%o", "%o", { "%j": "%j %j %%" })
debug("%o", "%o", { "%j": "%j %j %%" })
```

```
  foo '' { '%j': '%j %j %%' } 1 2 3 +0ms
  foo '' { '%j': '%j %j %%' } +4ms
[...]

PR #770, opened 2020-07-02
fix(common): escape replaced value

blastZ:
Something like this `log('I like %O', { name: '%sugar%' });` will cause output incorrect.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -99,6 +99,7 @@ function setup(env) {
 				if (typeof formatter === 'function') {
 					const val = args[index];
 					match = formatter.call(self, val);
+					match = match.replace(/%/g, () => '%%');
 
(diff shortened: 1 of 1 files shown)


Qix-:
Can you add tests for this, please? I'm not sure which sorts of things this fixes.

blastZ:
> Can you add tests for this, please? I'm not sure which sorts of things this fixes.

The default formatters will replace `%o` and `%O`, so when arguments go to the `log` function, `util.format` will get arguments like this `["...'%sugar%'... }", '...']`. So if you don't escape the `%` in the value, `util.format` will replace `%s` with the next argument. That's why issue #766 happend.

818Nawaf (approved):

PR #778, opened 2020-08-16
Fix only replacing wildcard at end in `toNamespace`

Anoesj:
The docs state that, when running `debug.disable()`, it will return the namespaces currently enabled and skipped. They are, however, often malformed. This seems to be because the `toNamespace` method only replaces a trailing `*`.

```javascript
debug.enable('api:*,-*:verbose,-api:*:verbose');
const prevConfig = debug.disable();
console.log(prevConfig); // -> api:*,-.*?:verbose,-api:.*?:verbose
```

`console.log(prevConfig);` should have returned `api:*,-*:verbose,-api:*:verbose`.

The issue seems to be fixed by removing the `$` in the replace regexp in `toNamespace`.

It might be nice to improve the following test too:
```javascript
describe('rebuild namespaces string (disable)', () => {
	it('handle names, skips, and wildcards', () => {
		debug.enable('test,abc*,-abc');
		const namespaces = debug.disable();
		assert.deepStrictEqual(namespaces, 'test,abc*,-abc');
		// ^ could use more tests, maybe: *,test,-test,abc*,*abc,abc:*:xyz,-abc:*:xyz,-*:verbose (etcetera)
	});
	// ...
});
```
However, for the sake of keeping this PR clean, I haven't added that to this PR.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -239,7 +239,7 @@ function setup(env) {
 	function toNamespace(regexp) {
 		return regexp.toString()
 			.substring(2, regexp.toString().length - 2)
-			.replace(/\.\*\?$/, '*');
+			.replace(/\.\*\?/, '*');
 	}
 
 	/**


src/common.js
```diff
@@ -239,7 +239,7 @@ function setup(env) {
 	function toNamespace(regexp) {
 		return regexp.toString()
 			.substring(2, regexp.toString().length - 2)
-			.replace(/\.\*\?$/, '*');
+			.replace(/\.\*\?/, '*');
```
curtisman, line 242: I think you need to set the regex to match globally, i.e. /\.\*\?/g

curtisman:
I wonder how to get the maintainer's attention to get contribution merged in?

Qix-:
I'm always looking, the project is currently blocked due to Node feature parity issues when moving from CJS to ESM. Please see the roadmap pinned on the repo.

Further, `.enable()` and `.disable()` have a ton of issues as it stands and need to be rethought/reworked entirely.

818Nawaf (approved):

PR #799, opened 2020-12-08
caching enabled status per logger

omg:
Can we have this to avoid rechecking rules on each call?

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -60,7 +60,8 @@ function setup(env) {
 	function createDebug(namespace) {
 		let prevTime;
 		let enableOverride = null;
-		let namespacesCache, enabledCache;
+		let namespacesCache;
+		let enabledCache;
 
 		function debug(...args) {
 			// Disabled?
@@ -123,14 +124,14 @@ function setup(env) {
 			configurable: false,
 			get: () => {
 				if (enableOverride !== null) {
-					return enableOverride
+					return enableOverride;
 				}
 				if (namespacesCache !== createDebug.namespaces) {
 					namespacesCache = createDebug.namespaces;
 					enabledCache = createDebug.enabled(namespace);
 				}
-				return enabledCache;
 
+				return enabledCache;
 			},
 			set: v => {
 				enableOverride = v;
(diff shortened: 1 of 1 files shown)


Qix-:
Please explain what this does. What problem are you facing?

omg:
Current implementation checks whether the namespace is enabled against the lists of regexes each time the debug instance is called. This change addresses this by caching the enabled flag in the debug instance to avoid reevaluating it each time. Since the library supports dynamically changing namespace configuration, the change also addresses this by caching reference to namespace config in the instance and comparing it against the current config (lazily). When the references are different the new config reference is replaces the previously cached one and the enabled flag reevaluated so nothing breaks but redundant regex checks are cut off.

src/common.js
```diff
@@ -120,7 +121,17 @@ function setup(env) {
 		Object.defineProperty(debug, 'enabled', {
 			enumerable: true,
 			configurable: false,
-			get: () => enableOverride === null ? createDebug.enabled(namespace) : enableOverride,
+			get: () => {
+				if (enableOverride !== null) {
+					return enableOverride
```
Qix-, line 126:
```suggestion
					return enableOverride;
```

src/common.js
```diff
@@ -120,7 +121,17 @@ function setup(env) {
...
+				}
+				if (namespacesCache !== createDebug.namespaces) {
+					namespacesCache = createDebug.namespaces;
+					enabledCache = createDebug.enabled(namespace);
+				}
+				return enabledCache;
+
```
Qix-, line 133:
```suggestion

				return enabledCache;
```

src/common.js
```diff
@@ -60,6 +60,7 @@ function setup(env) {
 	function createDebug(namespace) {
 		let prevTime;
 		let enableOverride = null;
+		let namespacesCache, enabledCache;
```
Qix-, line 63:
```suggestion
		let namespacesCache;
		let enabledCache;
```

Qix- (changes requested):
Sounds good, thanks!

Just a few small nitpicks and I'll merge.

Qix-:
Released as `4.3.2` on the `beta` tag.

diverdowngit:
juuhhyggbb

diverdowngit:
2.6.9

PR #828, opened 2021-05-21
Remove accidental epizeuxis

taylor1791:
While reading the source I discovered this sentence, emphasis mine: "the namespace string _for the_ _**for the**_ debug instance to be colored." Epizeuxes are an incredibly powerful technique, but this use seems to fall short of its potential. Also, it was probably an accident.


Qix- (approved):

Qix-:
Thanks :)

PR #852, opened 2021-10-14
Use ESModules instead of module.exports (#786)

### LINKED ISSUE #786: Use ESModules instead of module.exports
Today, debug use module.exports facility.
But nodejs and browser can support import / export notation, aka es6 modules.
While preparing the v5, why not going into that mode?
#656 ?

Konakona-chan:
Hello there.

As for 2021, NodeJS versions <12 without ESM support (are deprecated)[https://nodejs.org/en/about/releases/]. 

This PR switches `debug` package from CJS to ESM version, so it can be used in modern applications based on asynchronous execution.

Resolves #786.

Implementation notes:
1. This changeset updates dependencies to the latest ones to reduce package fragmentation and introduce ESM-based dependencies
2. It removes `debug.destroy()` function according to deprecation notice (notably, it updates `ms` to canary version from https://github.com/vercel/ms/issues/159)
3. It does not change API for existing CJS users who already used `import` statement; however `require(...)` should be replaced to `import` as described in `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ technique. Works in Node.js and web browsers.
 This package is [ESM only](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c):
 Node 12+ is needed to use it and it must be `import`ed instead of `require`d.
 
-[npm][]:
+[npm](https://www.npmjs.com/package/debug):
 
 ```sh
 npm install debug


Qix-:
This package is going to be rewritten in the next version, so there's not much I can do with this PR as it stands. Thank you for taking the time, however!

Konakona-chan:
@milahu , I was able to overcome the issue with Svelte+Vite and unified (esm-only) -> markdown -> debug (cjs-only) with `optimizeDeps`:
https://gitlab.com/Kona-chan/kona-chan.gitlab.io/-/blob/master/svelte.config.js#L32

FossPrime:
Please resubmit this.

Deno and UNPKG with ?module are broken because of this.

PR #858, opened 2021-12-30
Update version

merinpriyasha:
upgraded version is 2.6.9


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "debug",
-  "version": "1.0.5",
```
merinpriyasha, line 3: Update the version as 2.6.9

PR #866, opened 2022-02-02
Add section about configuring JS console to show debug messages

gitname:
In this branch, I added a section to the README in which I tell the reader how they can configure their JavaScript console to display messages sent to it via `console.debug()`.

Related issue: https://github.com/debug-js/debug/issues/808#issuecomment-1027529456

CC: @Qix-

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,13 @@ setInterval(function(){
 }, 1200);
 ```
 
+### Configuring the JavaScript console to show debug messages
+
+By default, `debug` uses the `console.debug()` function to send messages to the JavaScript console. That function sends messages to the JavaScript console at the "debug" log level. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/console/debug), some JavaScript consoles are—by default—configured to _hide_ log entries that are sent to it at the "debug" log level.
+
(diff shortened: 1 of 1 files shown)


Qix-:
This is a bit verbose, perhaps just adding the image (using [`this format`](https://www.digitalocean.com/community/tutorials/markdown-markdown-images) instead of an HTML tag) directly below the section about `localStorage` would be the best since that's where readers will be anyway.

Does that work?

gitname:
Hi @Qix- , that works for me. I'll make it more concise tomorrow (getting a little late in the day for me here).

The reason I used the HTML tag is that I saw that the other screenshots in the file were embedded using that same tag. It also allowed me to specify a width for the image.

gitname:
By the way, I'm also OK with you making changes to the branch directly, in case you want to wrap this up before I get to it tomorrow.

Qix-:
No worries, whenever works :) I'm not a computer right now anyway.

Good point about the image, feel free to keep it as a tag then.

gitname:
I rewrote it so it is more concise.

Qix- (approved):
Thanks!

PR #895, opened 2022-06-15
add namespace caching

KCErb:
I'd like to be able to extend a namespace at a per-function call level like so:

```javascript
class Foo {
  constructor() {
    this.log = debug('Foo');
    this.log('init');
  }
  bar() {
    const log = this.log.extend('bar');
    log('bar');
  }
}
```

The trouble is that each time I call bar, I reset the clock, I'd rather get my timestamp diffs between calls to the same namespace.

While writing this and updating tests I realized that one of the tests relied on a lack of caching between one instantiation of a `foobar` namespace and another. So I updated that to use `foobaz` but it doesn't really address the more important question of _is this breaking behavior_?

If it is, but the maintainers still agree this is valuable, perhaps there could be an option to enable caching? Or we can lift it to the user layer to overwrite extend like we do with formatters and colors? Just spitballing here....

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -146,9 +146,17 @@ function setup(env) {
 		return debug;
 	}
 
+	const debugCache = new Map();
 	function extend(namespace, delimiter) {
-		const newDebug = createDebug(this.namespace + (typeof delimiter === 'undefined' ? ':' : delimiter) + namespace);
+		const newNamespace = this.namespace +
+			(typeof delimiter === 'undefined' ? ':' : delimiter) +
+			namespace;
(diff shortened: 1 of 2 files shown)


Qix-:
We had namespace caching before and we faced a lot of users leaking memory (due to creating many different namespaces), so it was removed. It looks like this code is still vulnerable to that, just with `Map` instead of a `{}`.

This is most likely going to be handled at the higher level in the next major version (or at least whenever I am able to implement the user-handlers) so that people have more control over things without needing to worry about e.g. leaking the like.

Thanks for the PR though :)

PR #915, opened 2022-11-03
Exploratory PR for backwards-compatible changes & cross-env alignment: timeFormat

webketje:
I started this effort because I was in sore need of localized dates (we have Java & Node microservices and it's a pain that one takes into account timezone while the other does not). Then I realized there was a lot more I could do that would be both backwards-compatible, more powerful and more aligned across browser and Node. So this PR:

* Adds timeFormat option to debug instances, settable as DEBUG_TIME_FORMAT env var
* Deprecates hideDate option, equated to timeFormat === 'none'
* Allows timeFormat of choice: none, iso, diff, or localized (taking into account process.env.TZ in Node, and timezoneOffset in browser)
* Normalizes behavior across browser & Node: all timeFormats are available in both, but the defaults are kept as-is.
* Deprecates createDebug.humanize as public member and adds createDebug.withTimeFormat helper instead
* Documents this in the README.
* Adds debug format tests that are compatible with browser & Node.

@Qix- I was surprised at how outdated (dev)Dependencies are and was unable to run `npm test` (only `mocha test.js` directly).
The browserify build was not documented so I added an NPM script for it.. It may be useful to add it to the NPM files for those of us wanting to use debug in the browser by copying the dist to an asset folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ When actively developing an application it can be useful to see when the time sp
 
 <img width="647" src="https://user-images.githubusercontent.com/71256/29091486-fa38524c-7c37-11e7-895f-e7ec8e1039b6.png">
 
-When stdout is not a TTY, `Date#toISOString()` is used, making it more useful for logging the debug information as shown below:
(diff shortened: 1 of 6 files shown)


webketje:
@Qix feedback?

818Nawaf (approved):

PR #934, opened 2023-05-19
Fix/depth  Resolve debug: [object object] regular matching error in browser localStorage

wj0990:
It is not clear where this object is written. Cause this regular error

 in  common.js
```js
 if (namespaces[0] === '-') {
   createDebug.skips.push(new RegExp('^' + namespaces.slice(1) + '$'));
} else {
	createDebug.names.push(new RegExp('^' + namespaces + '$'));
// }
```
error

1521684460964_ pic

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "debug",
-  "version": "4.1.0",
+  "version": "4.2.0",
   "repository": {
     "type": "git",
     "url": "git://github.com/visionmedia/debug.git"
--- a/src/browser.js
+++ b/src/browser.js
@@ -189,8 +189,7 @@ function save(namespaces) {
 		if (namespaces) {
 			exports.storage.setItem('debug', namespaces);
 		} else {
-			exports.storage.removeItem('debug');
-		}
+			exports.storage.setItem('debug', typeof(namespaces) ==='string' ? namespaces : JSON.stringify(namespaces));		}
 	} catch (error) {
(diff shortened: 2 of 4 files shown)


Qix-:
Why are you passing an object to the namespaces?

wj0990:
> 为什么要将对象传递给名称空间？

We rely on a lot of third-party libraries. These large third-party libraries all depend on your package. I don’t know where it is written. I haven’t added it to my project.

Qix-:
You're using webpack. You should be able to turn on sourcemaps and get an accurate stacktrace as to where it's happening.

Qix-:
Sorry, but I'm not really interested in adding this support. The library is clearly being incorrectly used somewhere, you'll need to find where that's happening and open a PR against their codebase if it's happening outside yours.

wj0990:
> Sorry, but I'm not really interested in adding this support. The library is clearly being incorrectly used somewhere, you'll need to find where that's happening and open a PR against their codebase if it's happening outside yours.

Thanks, I tried it, but the development environment can't reproduce it
  It's a pity that I can't contribute. Several libraries in my current project dependencies depend on your package. There are more than a dozen projects of the same type nested in a desktop reference through the <ifarme> tag. The item has been in production for two years without any anomalies. The development environment has never been encountered before, that is, the recent production environment users access it from the desktop application. Several users online encountered this problem. I'm thinking of other ways.

PR #943, opened 2023-09-15
Enable logging on stderr or stdout depending of provided levels

bdelville:


Qix-:
Hi, thanks for the PR but I won't be accepting this - it breaks expectations as that's not how stdout or stderr are meant to be used.

PR #965, opened 2024-06-05
Add option to log to STDOUT

### LINKED ISSUE #964: Allow choice between STDERR and STDOUT
I have a situation where Atlassian Bamboo is being used for deployments (Don't blame me--not my choice). Unfortunately, in `log()` of `node.js` the file descriptor is hard coded to `stderr`. This causes everything to come out blood red in the logs. Not exactly the best thing to see in logs. 

With Bamboo the way to prevent this is to write the logs to `stdout` and then the output would be treated as normal log output and not display in red. 

Unfortunately, my knowledge level of NodeJS is not that great, but I think I just found how this could be triggered. Expect to see a PR in a few mins.

hickey:
This change recognizes the `DEBUG_USE_STDOUT` environmental variable to allow log messages to be directed to `STDOUT` when the NodeJS version is being used. Please refer to #964 for the use case to support this change. 

I have added a test that I think will work for testing this functionality. Since my NodeJS knowledge is not that great I can not say for certain that it works as I expect it to. It is my hope that you will correct any problem with the test. 

Closes #964


hickey:
Anybody still working on this project?

818Nawaf (approved):

endel:
I'm also looking forward to have this feature incorporated into `debug`. 

When using `debug` with PM2, all "debug" logs end up displayed in the "errors" log file, which is not intuitive, considering those are just debug logs, and not necessary error logs.

endel:
I just realized what I need is stated in the README, the solution is:

```javascript
const mydebug = debug('myapp:debug')
mydebug.log = console.debug.bind(console);
```

PR #969, opened 2024-07-26
Avoid using deprecated RegExp.$1

bluwy:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/n

`RegExp.$1` is deprecated and may be dropped support in the future. This PR refactors to avoid it.

I tested manually in Firefox and other browsers that it still works as before.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -125,6 +125,8 @@ function useColors() {
 		return false;
 	}
 
+	let m;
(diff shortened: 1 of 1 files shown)


Qix-:
Released as `4.3.6` :)

PR #973, opened 2024-09-19
remove 22kb readme, use link instead

KieranHolroyd:
Simple change, makes the unpacked size half of what it currently is while still having a link to the readme, which at 32,659,277 (rounding to 33) means this could save on the order of 0.6TB of internet traffic per week (34.4TB yearly).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "files": [
     "src",
     "LICENSE",
-    "README.md"
+    "readme-link"
   ],
   "author": "Josh Junon (https://github.com/qix-)",
   "contributors": [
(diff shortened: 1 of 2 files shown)


Qix-:
Hi, looks nice but I can't find any documentation on this. What is `readme-link`?

KieranHolroyd:
Sorry, I should have probably explained, it's simply a text file with a link to the readme, it's not any kind of standard, just a file that isn't any variant of actual readme files. This is a highly unimportant PR, but it's also so simple I thought it would be worth doing anyway.

_For context I was looking through the expressjs 5.0 release, looking at the changes, found this library and saw the number of weekly downloads, did some napkin maths and made these changes, I'll be honest, I didn't put a great deal of thought into it other than wanting to keep a versioned readme link in the package, while reducing the size_

Qix-:
Ah I see, yeah unfortunately that's not really feasible since the readme must be available on NPM for them to display it.

Thank you though!

PR #978, opened 2024-11-12
Fix node.js file.

DigitalWolf98:
This fix removes the error message require supports-color in esm (Node ver. 23.2.0).


Qix-:
This does not fix it.

PR #983, opened 2024-12-07
support inlined process.env.DEBUG

timvandam:
Some browser-like environments do not support localStorage nor process.env, but can be provided environment variables because they automatically get inlined (i.e. process.env.XYZ gets replaced by whatever value XYZ is during a build step). This change supports this scenario without breaking previous behavior.


Possumkitty (approved):
Thanks so much this really helped me!!

PR #984, opened 2024-12-13
fix: backslashes in template string didn't work after release 4.4.0

guillaumerxl:
Hello, 

Backslashes in template string is not taken into account after the changes made in the new `matchesTemplate` function from [4.4.0](https://github.com/debug-js/debug/releases/tag/4.4.0). 

When you need to escape a character (like `[` which is not valid as template string unless escaped), the debug is disabled because the character doesn't match the namespace. 

This PR is adding backslash as a special character

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -195,13 +195,15 @@ function setup(env) {
 		let starIndex = -1;
 		let matchIndex = 0;
 
+		const skippedChars = ['*', '\\']
+
 		while (searchIndex < search.length) {
-			if (templateIndex < template.length && (template[templateIndex] === search[searchIndex] || template[templateIndex] === '*')) {
-				// Match character or proceed with wildcard
-				if (template[templateIndex] === '*') {
+			if (templateIndex < template.length && (template[templateIndex] === search[searchIndex] || skippedChars.includes(template[templateIndex]))) {
+				// Match character or proceed with wildcard or backslash
+				if (skippedChars.includes(template[templateIndex])) {
 					starIndex = templateIndex;
 					matchIndex = searchIndex;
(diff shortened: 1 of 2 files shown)


Qix-:
This isn't the right solution I don't think. This will treat `\\` as a wildcard, which doesn't make sense. It needs to have its own special handling (whereby the next character is consumed directly, which right now the only case where that wouldn't happen is if it's a `*`).

Further, we should emit a deprecation notice as backslashes and other regex-related characters should _not_ have meaning in namespaces (it was never explicitly supported nor disallowed, but should have been strictly specified a long time ago).

I can work on this at some point but for now you should pin to the prior release.

ibc:
> Further, we should emit a deprecation notice as backslashes and other regex-related characters should _not_ have meaning in namespaces (it was never explicitly supported nor disallowed, but should have been strictly specified a long time ago).

The README literally says:

> ### Wildcards
> The `*` character may be used as a wildcard. Suppose for example your library has debuggers named "connect:bodyParser", "connect:compress", "connect:session", instead of listing all three with `DEBUG=connect:bodyParser,connect:compress,connect:session`, you may simply do `DEBUG=connect:*`, or to run everything using this module simply use `DEBUG=*`.

I agree that the README says nothing about backslashes. However it does say things about wildcards. And the problem is that wildcards support has been **completely removed** in 4.4.0 when using `debug.enable()` API, **despite it was documented and still it is documented in the README**:

https://github.com/debug-js/debug/issues/985

Qix-:
Wildcards have not been "completely removed", please check the commits leading up to the release. I spent a lot of time re-implementing them using a manual parser. If they aren't working, please post a **new issue** with your issue.

PR #1018, opened 2025-11-07
chore: inlining `ms`

hyperz111:
We **ONLY** use `ms` for converting millisecond number to human-readable string. So, we can inline it for **smaller** install size.

**Edit:** `debug` use ±15% code of `ms@2`.


hyperz111:
cc @Qix-

PR #1035, opened 2026-05-25
feat: add option to disable printf-style formatting

### LINKED ISSUE #976: Way to disable printf-style feature temporarily
We are seeing some interesting side-effect of the [printf-style feature](https://github.com/debug-js/debug?tab=readme-ov-file#formatters) outlined here: https://github.com/sequelize/sequelize/issues/17448

Long story short, we are using debug to print SQL queries that are executed against database servers and whenever people use a like or ilike query in combination with % at the beginning, it sometimes happens that the logs show `NaN`. This happens when the search term starts with any of the printf-style characters, e.g. for something like `SELECT * FROM foo WHERE name LIKE %fauzi%`. The logs will then contain `NaNauzi`.

Would it be possible to force a debug log statement to ignore any printf formatting?

Something like:

```
const log = debug('sql')

log("%fauzi%", { printfFormatting: false})
```

wyf027:
## Summary

- Adds per-call `{ printfFormatting: false }` option and instance-level `debug.printfFormatting = false`
- Escapes `%` in log arguments before downstream printf handling so strings like `LIKE %fauzi%` no longer produce `NaN` when extra args are present
- Keeps existing debug formatters (`%o`, `%O`, `%j`) behavior unchanged

## Test plan

- [x] `npm run test:node` (19 tests)
- [x] Regression test for SQL-like `%f` strings with extra args

Closes #976

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -50,6 +50,34 @@ function setup(env) {
 	}
 	createDebug.selectColor = selectColor;
 
+	function isPrintfOptions(value) {
+		return (
+			value !== null &&
+			typeof value === 'object' &&
+			!Array.isArray(value) &&
+			Object.prototype.hasOwnProperty.call(value, 'printfFormatting')
+		);
+	}
+
+	function resolvePrintfFormatting(args, defaultValue) {
+		const last = args[args.length - 1];
+
+		if (!isPrintfOptions(last)) {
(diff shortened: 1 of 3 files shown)


Qix-:
Please stop carpet bombing repositories with low-effort PRs.

PR #1043, opened 2026-07-12
docs: clarify log precedence

### LINKED ISSUE #873: Can't set global `log()` function overriding namespaces ones
On the example located at https://github.com/debug-js/debug#output-streams, it shows that's possible to set custom `log()` functions for each namespace, but also that setting it directly on the `debug` module takes precedence and will be used instead of the namespace ones. This is specially useful in my use case, since I want to capture [Mediasoup](https://mediasoup.org/) debuging [Logger](https://github.com/versatica/mediasoup/blob/v3/node/src/Logger.ts] and redirect it to [Moleculer logger](https://github.com/moleculerjs/moleculer/blob/master/src/loggers/base.js). Problem is that according to https://github.com/debug-js/debug/blob/c0805cc4d3b2c7c987567e99ecaec78840516d7b/src/common.js#L112-L113 namespaces `log()` function has absolute priority, so it's not possible to do it globally, and it's not possible to directly bind on top of Mediasoup Loggers since they are private, so I'm in an end road.

Since the linken example at https://github.com/debug-js/debug/blob/master/examples/node/stdout.js doesn't exists too, I'm not sure if the documentation is outdated or wrong, or if setting `debug.log()` should effectively override per-namespace `log()` function and this is in fact a bug. Besides fixing this one way or another, how can I solve my problem of redirect Mediasoup log messages to Moleculer?

vibhor-aggr:
## Summary

- clarify the README output-streams section for `debug.log` precedence
- state that a namespace-specific `log` method takes precedence over the global `debug.log`
- update the example comments/output text to match the current implementation

Closes #873.

## Verification

- smoke check proving namespace `log` wins over global `debug.log`
- `npm run test:node`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,9 @@ In Chromium-based web browsers (e.g. Brave, Chrome, and Electron), the JavaScrip
 
 ## Output streams
 
-  By default `debug` will log to stderr, however this can be configured per-namespace by overriding the `log` method:
+By default `debug` will log to stderr. This can be configured globally by
+overriding `debug.log`, or per namespace by overriding the namespace's `log`
+method. A per-namespace `log` method takes precedence over `debug.log`.
 
 Example _stdout.js_:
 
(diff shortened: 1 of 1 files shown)


Qix-:
This doesn't add anything.

PR #1044, opened 2026-07-12
docs: note empty DEBUG behavior

### LINKED ISSUE #845: Undocumented ```delete process.env.DEBUG``` when variable is empty
I would just like to point out to the fact that when required, debug package will remove the DEBUG env variable if empty.

I think it would be nice if this is mentioned in the readme, as I feel its a non obvious behavior.

The use case I noticed this on is:

### command:
`npm install dotenv`
`DEBUG= npx mocha --require bootstrap.js`

### files:
.env
`DEBUG=*`

bootstrap.js
`console.log(process.env.DEBUG);`
`require('dotenv').config();`
`console.log(process.env.DEBUG);`

### output:
`undefined // we lost the DEBUG env already`
`'*' // we get .env file DEBUG setting, but we should have gotten '', because dotenv never overrides set env variables (* actually after version 4 I believe)
`
### problem:
mocha uses the debug package which deletes the env variable. Our user code in bootstrap.js then has no way of knowing if DEBUG was set to an empty string or not supplied. Therefore we get full debug info instead of none.

More general - any code used before using process.env.DEBUG might delete the env variable through just depending on the debug package.

### solution

I propose we either not delete the DEBUG variable or have a section in documentation to explain and suggest maybe using  `DEBUG=,` to disable debug logging.

vibhor-aggr:
## Summary

- document that an empty `DEBUG` value is treated as disabling all namespaces
- note that saving this disabled state removes `process.env.DEBUG`
- mention `DEBUG=,` as a non-empty disabled value when other code needs to observe that `DEBUG` was set

Closes #845.

## Verification

- smoke check proving `DEBUG=` is removed after requiring `debug`
- `npm run test:node`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,10 @@ change the behavior of the debug logging:
 | `DEBUG_DEPTH` | Object inspection depth.                    |
 | `DEBUG_SHOW_HIDDEN` | Shows hidden properties on inspected objects. |
 
+If `DEBUG` is set to an empty value, `debug` treats it the same as disabling all
+namespaces and removes `process.env.DEBUG` when it saves that disabled state.
+Use a non-empty disabled value such as `DEBUG=,` when other code must still be
+able to observe that the environment variable was set.
 
 __Note:__ The environment variables beginning with `DEBUG_` end up being
 converted into an Options object that gets used with `%o`/`%O` formatters.


Qix-:
I don't really appreciate the AI low-effort PRs.

PR #1050, opened 2026-08-14
fix: merge namespaces in enable() instead of flushing them

### LINKED ISSUE #425: `debug.enable()` flushes enabled namespaces

arronKler:
`debug.enable()` currently clears `names`/`skips` before applying the new list. That means previously enabled namespaces (including those from `DEBUG`) disappear:

```js
DEBUG=foo node -e 'var dbg = require("debug"); dbg.enable("bar"); console.log(dbg.enabled("foo"))'
// false
```

This also breaks independent `enable()` calls from different modules that share one `debug` install.

**Solution:** treat `enable()` as additive. Incoming namespaces are merged with the existing set (skip tokens still disable a matching name; enabling a name again removes its skip). `disable()` still clears everything and is the way to replace:

```js
debug.disable();
debug.enable('bar');
```

Fixes #425

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,11 +313,19 @@ Usage :
 `enable(namespaces)`  
 `namespaces` can include modes separated by a colon and wildcards.
    
-Note that calling `enable()` completely overrides previously set DEBUG variable : 
+Calling `enable()` merges with previously set namespaces (including `DEBUG`)
+instead of replacing them:
 
 ```
-$ DEBUG=foo node -e 'var dbg = require("debug"); dbg.enable("bar"); console.log(dbg.enabled("foo"))'
-=> false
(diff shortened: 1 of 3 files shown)


official-burak (commented):
enable() is documented as a full replace, including the DEBUG=foo example this PR rewrites. Code that calls debug.enable('production') today relies on that to drop whatever DEBUG was. Making it additive would keep those namespaces on, which is a behavior change rather than a bugfix.

If additive enable is useful, it needs a new method or an opt-in, not a silent change to enable().

PR #1052, opened 2026-08-17
docs: drop dead wzrd.in standalone browser build link

JiuYue0820:
## Problem

`https://wzrd.in/standalone/debug@latest` now returns HTTP 404 on GET. The README still advertised that hosted browserify-as-a-service build.

`https://github.com/substack/node-browserify` still works, but the project lives at `https://github.com/browserify/browserify`.

## Change

- Remove the dead wzrd.in standalone-build promise.
- Keep the local browserify instructions and point that repo at its current org URL.

Verified with `curl -L GET` on 2026-08-17:

- `https://wzrd.in/standalone/debug@latest` -> 404
- `https://github.com/browserify/browserify` -> 200

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,8 +214,8 @@ debug('this is hex: %h', new Buffer('hello world'))
 
 ## Browser Support
 
-You can build a browser-ready script using [browserify](https://github.com/substack/node-browserify),
-or just use the [browserify-as-a-service](https://wzrd.in/) [build](https://wzrd.in/standalone/debug@latest),
+You can build a browser-ready script using [browserify](https://github.com/browserify/browserify),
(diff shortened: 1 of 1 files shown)


lbesecker195 (approved):
Checked the links: `https://wzrd.in/standalone/debug@latest` returns 404 (the wzrd.in landing page still loads, but the standalone builds it generates don't), and `github.com/substack/node-browserify` redirects to `github.com/browserify/browserify`, so pointing at the canonical repo and dropping the build-as-a-service line is right. LGTM.

Qix-:
The whiplash I got from hovering over the "report" button after reading "drop dead" is indescribable :D unfortunately not the first death threat I would have gotten over this package.

Anyway, thanks!

